PR #146, opened 2014-07-24
"What You See Is What You Get"-export for volumina's ImageView2D

christophdecker:
Adds a "camera"-button to each ImageView2D HUD allowing the user to export (a custom ROI of) the scene as he/she sees it (with all semi-transparent overlays, objects, etc.) to a stack of images. 

(Tested under Ubuntu 14.04 for several ilastik applets, including whole Pixel Classification and Object Classification workflow. Would be good if @stuarteberg and @czhci could test this under Mac OS and Windows once added to master.)


stuarteberg:
Superseded by #155

PR #152, opened 2015-02-03
Masked arrays i.e. Handle frames of variable size (with known maximum)

jakirkham:
Provides support for masked arrays in volumina. In most cases, this is a series of minor fixes to volumina to ensure support of masked arrays. In a few cases, more heavy duty code is required to bring about support. Most of the support is already available to thanks to ( https://github.com/hmeine/qimage2ndarray ). Masked arrays appear basically the same NumPy ndarrays except that they mask results. In an image source, the masked regions are made transparent. This is related to several issues ( https://github.com/ilastik/ilastik/issues/1076 ) (  https://github.com/ilastik/lazyflow/issues/162 ) ( https://github.com/ilastik/volumina/issues/151 ).

### DIFF
--- a/volumina/layer.py
+++ b/volumina/layer.py
@@ -185,6 +185,7 @@ class Layer( QObject ):
         self._toolTip = ""
         self._cleaned_up = False
 
+        self._updateNumberOfChannels()
         for datasource in filter(None, self._datasources):
             datasource.numberOfChannelsChanged.connect( self._updateNumberOfChannels )
 
@@ -203,9 +204,22 @@ class Layer( QObject ):
         self.contexts = []
 
     def _updateNumberOfChannels(self):
-        newchannels = self._datasources[0].numberOfChannels
-        for datasource in self._datasources[1:]:
-            newchannels = min( datasource.numberOfChannels, newchannels )
+        # As there can many datasources and they can all be None,
+        # grab numberOfChannels for those that are defined.
+        # That is, if there are any datasources.
+        newchannels = []
+        for i in xrange(len(self._datasources)):
+            if self._datasources[i] is not None:
+                newchannels.append(self._datasources[i].numberOfChannels)
+
+        # If the datasources are all None or there aren't any,
+        # default to 1 channel as we must have at least 1.
+        if not newchannels:
+            newchannels = [1]
+
+        # Get the smallest number of channels that works for all.
+        newchannels = min(newchannels)
+
         if newchannels != self.numberOfChannels:
             # Update property (emits signal)
             self.numberOfChannels = newchannels
--- a/volumina/layer.py
+++ b/volumina/layer.py
@@ -185,7 +185,6 @@ class Layer( QObject ):
         self._toolTip = ""
         self._cleaned_up = False
 
-        self._updateNumberOfChannels()
         for datasource in filter(None, self._datasources):
             datasource.numberOfChannelsChanged.connect( self._updateNumberOfChannels )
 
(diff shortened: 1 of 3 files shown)


jakirkham:
A complimentary pull request for lazyflow can be seen as well ( https://github.com/ilastik/lazyflow/pull/163 ).

volumina/pixelpipeline/imagesources.py
```diff
@@ -420,8 +423,95 @@ def toImage( self ):
                 #FIXME: maybe this should be done in a better way using an operator before the colortable request which properly handles 
                 #this problem 
                 warnings.warn("Data for colortable layers cannot be float, casting",RuntimeWarning)
-                a=a.astype(np.int32)
-            vigra.colors.applyColortable(a, self._colorTable, byte_view(img))
+                a = np.asanyarray(a, dtype=np.uint32)
+
+            # If we have a masked array with a non-trivial mask, ensure that mask is made transparent.
+            _colorTable = self._colorTable
+            if np.ma.is_masked(a):
```
stuarteberg: This whole `if` statement is way too long and deeply nested.  Please break it into smaller functions so it's easier to see the big picture.

volumina/layer.py
```diff
@@ -203,9 +204,22 @@ def __init__( self, datasources, direct=False ):
...
+
+        # If the datasources are all None or there aren't any,
+        # default to 1 channel as we must have at least 1.
+        if not newchannels:
+            newchannels = [1]
+
+        # Get the smallest number of channels that works for all.
+        newchannels = min(newchannels)
+
         if newchannels != self.numberOfChannels:
```
stuarteberg: As far as I can tell, all of these changes in `layer.py` are solving a different problem than the focus of this PR.  It would be better if these changes could be broken out into their own PR.

stuarteberg:
This is good work -- thanks.  And thanks for cross-referencing this PR with the corresponding issues in `ilastik` and `lazyflow`.  Finally, thanks for writing good tests (especially for `AlphaModulatedImageSource`, which was lacking test coverage in the first place).

In the diffs of this PR, I made a couple comments requesting changes.  This can be merged after those are addressed.

volumina/pixelpipeline/imagesources.py
jakirkham: You're right. I will take a look at refactoring this.

volumina/layer.py
jakirkham: Initially, the goal of the issue was to handle imagestacks with variable shapes. In an attempt to handle the variable channel case, this was kept. However, this has since become a minor aspect of this PR and I will split it out into a separate PR.

jakirkham:
Thanks again for reviewing this. I responded to your comments and will work on some adjustments.

volumina/pixelpipeline/imagesources.py
jakirkham: I was thinking we could have a utility function that ensures there is a transparent layer as the first element of a colortable (adding one if it is missing). Where would you like me to put this utility function?

stuarteberg: Is that the right approach?  Some layers do not need (and should not have) a transparent layer.  If the color table is modified, the layer data would need to change to compensate for that.

But to answer your question: general utility functions that don't rely on other parts of volumina can go in the `volumina.utility` package.  If the function you have in mind somehow relies on functionality in `layer.py` or `imagesources.py`, then put it in one of those two files.

jakirkham: So, I think something got lost in transmission. Maybe this won't be totally clear until I submit changes, but here is my attempt to clear it up in writing.

First, as a utility function, it will be an optional piece of code that people can chose to use if they wish. It will not be something happening behind the scenes without anyone’s control.

Second, the only place I am planning to use this utility function is for ensuring the masked part of the array appears transparent, in other words, in place of lines 431-512. In this case, the case of masked arrays, a transparent layer is absolutely necessary. Further, this would leave other unmasked layers (i.e. ColortableImageLayer instances) untouched.

If something is still not clear, I'll blame it on that long portion of code and we can try again once we have a better piece of code to look at.

jakirkham: It turned out there was quite a bit of filler. Also, I dropped out handling the case of extremely large label images. This case seems impractical in terms of memory and possibly processor efficiency. Please let me know your thoughts.

jakirkham: On a side note, do you know what the max allowable type for the label image is for `vigra.colors.applyColortable`. It seems to be related to the function `pythonApplyColortable` ( https://github.com/ukoethe/vigra/blob/4510b643cbf1d62d1b9111c53236919eb1c2d063/vigranumpy/src/core/colors.cxx#L302 ), but the type isn't obvious. I have assumed uint32, which seem to work, but if it is uint64 I wouldn't want to unnecessarily disallow something.

jakirkham:
Made some improvements. Please let me know your thoughts.

jakirkham:
@stuarteberg, can I get some feedback on this?

jakirkham:
Thanks.

PR #155, opened 2015-03-26
Wysiwyg-Exporter number two

nielsbuwen:
This is an updated version of @christophdecker 's pull request #146 .
It fixes the following issues:
* Identical images won't be exported anymore https://github.com/ilastik/volumina/pull/146#issuecomment-50779070
* Random background is removed
* The ProgressDialog finishes in the "complete" state
* Camera tooltip changed to "Export Current Composite View"
* Indices are zero padded https://github.com/ilastik/volumina/pull/146#issuecomment-51498466
* Cancel button does respond https://github.com/ilastik/volumina/pull/146#issuecomment-51500248
* Stack size description is calculated correctly https://github.com/ilastik/volumina/pull/146#issuecomment-52197215
* Top right view axis order is now correct

* With the optional dependency "Wand" ( pip install Wand ) 3D stacks can be combined when exporting *.tif *.tiff and *.ico


stuarteberg:
This is very cool.  So far, I see only one issue: For some reason, the "X" view image doesn't seem to be exported with the correct dimensions.  The ROI seems to be flipped in X and Y.  Presumably, that's because its axes are [swapped_default](https://github.com/ilastik/volumina/blob/cf141a2092deeb1e28234a8b1e85641d6b5cbafb/volumina/volumeEditor.py#L160) (unlike the other two views).

For example, see the following screenshot, and the resulting exported image.  The dimensions are wrong.

screen shot 2015-03-26 at 1 44 43 pm
hi

volumina/imageScene2D.py
```diff
@@ -425,7 +425,7 @@ def drawBackground(self, painter, sceneRectF):
             for through in self._bowWave(self._n_preemptive):
                 self._tileProvider.prefetch(sceneRectF, through)
 
-    def joinRenderingAllTiles(self, viewport_only=True):
+    def joinRenderingAllTiles(self, viewport_only=True, rect=None):
         """
```
stuarteberg: Now you're using this function for non-test purposes.  That's fine, but please update the docstring here.

nielsbuwen:
The roi is now calculated correctly

inline-fixyz
exportyz

nielsbuwen:
I tested every possible combination of rotating and swapping: They all worked for me.

stuarteberg:
Perfect.  Nice work!

PR #159, opened 2015-08-10
Made `ui` a package as apparently it is treated as one in `setup.py`

jakirkham:
This is where it is treated as a package ( https://github.com/ilastik/volumina/blob/master/setup.py#L29 ).


jakirkham:
Failure appears to be due to an inability to download `tifffile` not an actual error with the code change.

stuarteberg:
That directory contains no `.py` files; just `.ui` files.  Maybe the right thing to do is to just remove the offending line from `setup.py`?

jakirkham:
Probably so. However, I don't know if there was a reason for it to be imported in the first place.

jakirkham:
It may be that this is required for it to be copied over to site-packages in the first place.

stuarteberg:
Hmm.  My guess is that the first entry in `package_data` should be deleted. (After all, `widgets.ui` isn't a package, as you pointed out.).  Instead, you should add `ui/*.ui` to the list of package data for the [`widgets` package](https://github.com/ilastik/volumina/blob/master/setup.py#L35).

jakirkham:
This appears to install the `.ui` files correctly.

stuarteberg:
OK, cool.  Thanks for giving it a try.

PR #160, opened 2016-01-29
Learning - using croppingMarkers for selecting training regions

ilastikdev:
Required for crop selection and training applets in the new Structured Learning Tracking workflow.

Changes facilitate selecting crops (regions of interest) in which we will train for tracking.

Also adds a TimeSlider bar, "Start Time", and "End Time" buttons
They act on a crop shape if in Cropping or Training applet and on dataShape otherwise.
Facilitated by new _crop_times in CropExtentsModel.

There is a known problem in master code from which croppingMarkers do not get properly mapped if rotation/reflexion buttons are used.


stuarteberg:
Would you mind adding a little bit more detail to the PR description above?  I need a little help understanding what functionality you added here.

volumina.egg-info/PKG-INFO
```diff
@@ -0,0 +1,10 @@
+Metadata-Version: 1.0
+Name: volumina
+Version: 0.6a
+Summary: Volume Slicing and Editing
+Home-page: https://github.com/Ilastik/volumina
+Author: UNKNOWN
+Author-email: UNKNOWN
+License: UNKNOWN
+Description: UNKNOWN
+Platform: UNKNOWN
```
stuarteberg: We don't include build artifacts (e.g. .egg-info files) into the repo.

volumina/croppingMarkers.py
```diff
@@ -47,12 +65,13 @@ def set_roi_3d(self, roi):
         # Set the extents as a roi
         self.set_crop_extents( zip( *roi ) )
 
-    def set_volume_shape_3d(self, shape3d):
-        # Since the volume size changed, 
+    def set_volume_shape_3d(self, starts, stops):
+        # Since the volume size changed,
```
stuarteberg: Since you changed this function signature to accept a roi instead of a shape, I think you should also change the function name.

volumina/croppingMarkers.py
```diff
@@ -65,21 +84,35 @@ def set_crop_extents(self, crop_extents):
...
+
+    def __init__(self, scene, axis, crop_extents_model, editable=True):
+
+        self._cropColor = Qt.white
+        self._editable = editable
 
-    def __init__(self, scene, axis, crop_extents_model):
         QGraphicsItem.__init__(self, scene=scene)
-        self.setFlag(QGraphicsItem.ItemHasNoContents);
-
```
stuarteberg: I might be missing something here, but I don't understand why you remove the `ItemHasNoContents` flag here.  Below, you provided an implementation of `paint()`, so why not set the flag here?

volumina/croppingMarkers.py
```diff
@@ -103,28 +141,270 @@ def dataShape(self, shape2D):
...
-        axis_3d = axislookup[self.axis].index( direction[0])
+
+        axis_3d = axislookup[self.axis].index(direction[0])
         crop_extents_3d = self.crop_extents_model.crop_extents()
+
         crop_extents_3d[axis_3d][index] = int(new_position)
         self.crop_extents_model.set_crop_extents( crop_extents_3d )
 
+    def mousePressEvent(self, event):
+
```
stuarteberg: Sure this function could be implemented more succinctly....

volumina/imageView2D.py
```diff
@@ -166,6 +171,48 @@ def __init__(self, parent, cropModel, imagescene2d):
         # For screen recording BlankCursor doesn't work
         #self.hiddenCursor = QCursor(Qt.ArrowCursor)
 
+    def _onRotateLeft(self):
+        print "-----------------------------------------------------------------------------------------------------"
```
stuarteberg: Eliminate all print statements, please.  Feel free to use `logger.debug()` instead.

volumina/imageView2D.py
```diff
@@ -84,6 +86,9 @@ def hud(self, hud):
 
         scene.axesChanged.connect(hud.setAxes)
 
+        #hud.rotLeftButtonClicked.connect(self._onRotateLeft)
+        #hud.rotRightButtonClicked.connect(self._onRotateRight)
+        #hud.swapAxesButtonClicked.connect(self._onSwapAxes)
```
stuarteberg: Remove comment cruft like this.

volumina/imageView2D.py
stuarteberg: Actually, it looks like this function and the two below are unused.  Just remove them entirely.

volumina/sliceSelectorHud.py
```diff
@@ -490,11 +505,111 @@ def createQuadViewStatusBar(self,
...
+    def _onTimeStartButtonClicked(self):
+        try: # cropping applet
+            self.timeSpinBox.setValue(
+                self.parent().parent().parent().parent().topLevelOperatorView.MinValueT.value) # opCropSelection
+        except:
+            try: # annotations applet
+                self.timeSpinBox.setValue(
+                    self.parent().parent().parent().parent().topLevelOperatorView.Crops.value[
+                        self.parent().parent().parent().parent()._drawer.cropListModel[
+                            self.parent().parent().parent().parent()._selectedRow ].name ]["time"][0])
```
stuarteberg: :dizzy_face: :scream: :cry:  :confounded: :sob:

volumina/sliceSelectorHud.py
```diff
@@ -502,11 +617,18 @@ def _handlePositionBoxValueChanged(self, axis, value):
         new_position[changed_axis] = value
         self.positionChanged.emit(*new_position)
 
-    def updateShape5D(self, shape5D):
-        self.timeSpinBox.setMaximum(shape5D[0]-1)
-        self.xSpinBox.setMaximum(shape5D[1]-1)
-        self.ySpinBox.setMaximum(shape5D[2]-1)
-        self.zSpinBox.setMaximum(shape5D[3]-1)
+    def updateShape5D(self, shape5Dmax, shape5DcropMin):
```
stuarteberg: I'd say this function should be renamed now that it takes more than just a shape.  Also, a subtle point: most functions that take a roi take the args in (min, max) order, not (max, min) order.  It will be easier to understand if we stick with that convention.

volumina/sliceSelectorHud.py
ilastikdev: A hack done very much in a hurry the day before our annual sfb1129 appearance.
"TimeStart" button is essential for efficiently demoing the training 
(essentially embedded ManualTrackingApplet)

I expected my second solution to show up here, but it obviously got lost in a stash...
will update.

volumina/croppingMarkers.py
ilastikdev: with distances for which you need multiplications

ilastikdev:
ilastik/volumina learning is ready to be pulled into
ilastik/volumina master

volumina/croppingMarkers.py
```diff
@@ -65,21 +133,41 @@ def set_crop_extents(self, crop_extents):
...
 class CroppingMarkers( QGraphicsItem ):
     PEN_THICKNESS = 1
 
     def boundingRect(self):
-        # Return an empty rect to indicate 'no content'
-        # This 'item' is merely a parent node for child items
-        return QRectF()
+        width, height = self.dataShape
+        return QRectF(0.0, 0.0, width, height)
+
```
stuarteberg: I think this isn't right.  Since this item is marked with the `ItemHasNoContents` flag, you should leave this implementation of `boundingRect()` unchanged (i.e. return an empty `QRect`).  And below, you should delete the implementation of `paint()` entirely, instead of reimplementing it as `def paint(...): pass`.

volumina/imageView2D.py
```diff
@@ -23,6 +23,7 @@
 from PyQt4.QtGui import QCursor, QGraphicsView, QPainter, QVBoxLayout, QApplication, QImage
 
 import numpy
+import threading
```
stuarteberg: unused import?

volumina/imageView2D.py
```diff
@@ -137,14 +138,14 @@ def __init__(self, parent, cropModel, imagescene2d):
         self._crossHairCursor = CrossHairCursor(self.scene())
         self._crossHairCursor.setZValue(99)
 
-        posModel = self.scene()._posModel
-        axis = self.scene()._along[1] - 1 # axis is 0,1,2 for X,Y,Z
-        self._sliceIntersectionMarker = SliceIntersectionMarker(self.scene(), axis, posModel)
+        self.posModel = self.scene()._posModel
+        self.axis = self.scene()._along[1] - 1 # axis is 0,1,2 for X,Y,Z
+        self._sliceIntersectionMarker = SliceIntersectionMarker(self.scene(), self.axis, self.posModel)
```
stuarteberg: I don't understand why it's necessary to add these as members to `self`.

stuarteberg: OK, I get it.  You're referencing them within `sliceSelectorHud.py`.  I guess that's fine.

volumina/positionModel.py
```diff
@@ -83,13 +83,13 @@ def __init__(self, parent=None):
         self._time       = 0
         self._channel    = 0
         self._shape5D    = [0,0,0,0,0]
-        
+        self._parent = parent
+
```
stuarteberg: I think it's preferable not to store `parent` as a member here in python, because it is already stored in the `QObject`, and accessible via the `parent()` accessor function.  (Avoiding this references reduces a cycle in the garbage collector, which is also nice.)

volumina/sliceSelectorHud.py
```diff
@@ -168,7 +168,7 @@ def __init__(self, parentView, backgroundColor, foregroundColor,
         font.setPixelSize(fontSize)
         self.spinBox.setFont(font)
         self.do_draw()
-
+        self._parentView = parentView
     def do_draw(self):
```
stuarteberg: Similarly, I think we should avoid storing this `parentView` member here.  (But in this case, the previous author forgot to pass `parent` to the base class (see above), so you'll need to fix that.)

volumina/croppingMarkers.py
stuarteberg: I still don't understand your comment about multiplications, but I understand that rewriting this function isn't a priority right now.  Just add a detailed docstring to this function and the one below.

ilastikdev:
cleaned up parents and opas

also changed the gui corner selection (with *)

stuarteberg:
OK, thanks.  Looking good now, as far as I can tell.  I'll do a bit of testing and merge this later today.

PR #162, opened 2016-02-24
hides time slider with start/end buttons for data with no time di…

ilastikdev:
…mension

and display tooltip for all three according to default (view)

- in cropping start/end buttons act on time crop boundaries
- in training also slider reacts in the crop time range

### DIFF
--- a/volumina/sliceSelectorHud.py
+++ b/volumina/sliceSelectorHud.py
@@ -469,10 +469,10 @@ class QuadStatusBar(QHBoxLayout):
     def setToolTipTimeButtonsCrop(self,croppingFlag=False):
         if croppingFlag==True:
             self.timeStartButton.setToolTip("Set the time coordinate to the beginning of the current crop.")
-            self.timeEndButton.setToolTip("Set the time coordinate to the beginning of the current crop.")
+            self.timeEndButton.setToolTip("Set the time coordinate to the end of the current crop.")
         else:
             self.timeStartButton.setToolTip("Set the time coordinate to the beginning of the current view.")
-            self.timeEndButton.setToolTip("Set the time coordinate to the beginning of the current view.")
+            self.timeEndButton.setToolTip("Set the time coordinate to the end of the current view.")
 
     def setToolTipTimeSliderCrop(self,croppingFlag=False):
         if croppingFlag==True:
--- a/volumina/sliceSelectorHud.py
+++ b/volumina/sliceSelectorHud.py
@@ -461,6 +461,25 @@ class QuadStatusBar(QHBoxLayout):
         self.zLabel.setHidden(False)
         self.zSpinBox.setHidden(False)
     
+    def hideTimeSlider(self,flag):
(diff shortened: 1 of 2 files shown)


volumina/sliceSelectorHud.py
```diff
@@ -461,6 +461,25 @@ def showXYZCoordinates(self):
...
+        self.timeEndButton.setHidden(flag)
+        self.timeStartButton.setHidden(flag)
+
+    def setToolTipTimeButtonsCrop(self,croppingFlag=False):
+        if croppingFlag==True:
+            self.timeStartButton.setToolTip("Set the time coordinate to the beginning of the current crop.")
+            self.timeEndButton.setToolTip("Set the time coordinate to the end of the current crop.")
+        else:
+            self.timeStartButton.setToolTip("Set the time coordinate to the beginning of the current view.")
+            self.timeEndButton.setToolTip("Set the time coordinate to the end of the current view.")
```
akreshuk: Could you call it "current dataset", not "current view"? In all the tooltips.

volumina/sliceSelectorHud.py
```diff
@@ -517,12 +536,15 @@ def createQuadViewStatusBar(self,
         self.timeSlider = QSlider(Qt.Horizontal)
         self.timeSlider.setMinimumWidth(10)
         self.timeSlider.setMaximumWidth(200)
+        self.timeSlider.setToolTip("Choose the time coordinate of the current view.")
```
akreshuk: Why not call the function you just defined (setToolTipTimeSliderCrop) for this? It will be easier to change the wording in future if it's all in one place.

volumina/volumeEditorWidget.py
```diff
@@ -245,6 +245,12 @@ def onShapeChanged():
                 cropMin = (self.editor.posModel.time,self.editor.cropModel._crop_extents[0][0],self.editor.cropModel._crop_extents[1][0],self.editor.cropModel._crop_extents[2][0],0)
                 self.quadViewStatusBar.updateShape5Dcropped(cropMin,self.editor.posModel.shape5D)
 
+            timeDim = self.editor.posModel.shape5D[0]
```
akreshuk: Why is this here and not in the _setupVolumeExtent()? Is it because you want to allow for crops with no time to be selected from volumes with time?

ilastikdev: New code deals with changes of data with T and data without T.
It is in the same block of code as:

"            singletonDims = filter( lambda (i,dim): dim == 1, enumerate(self.editor.posModel.shape5D[1:4]) )
..."
that deals with changes of XY vs. XYZ data.

Having in mind that _setupVolumeExtent() is only called from this line we could merge the whole proc, or leave it as it is with a minimal change of code base.

akreshuk: I'd rather have both the slider and the time spin box enabled/disabled in the same function, whichever that is.

PR #170, opened 2016-09-08
Py2to3 volumina

njase:
**Tag**: py2to3_v0.2
**Details**: PEP238 support for integer division using future import division.
**Test status**:
- Most of the changes tested - list attached ([v1.2_Overall_result.xlsx](https://github.com/ilastik/volumina/files/461749/v1.2_Overall_result.xlsx))
 - unit tests  
 - manual invocation of application; and/or 
 - python -3 switch
- Some changes could not be reached and are asserted <br \>
- Testing of Automatic tracking is pending and shall be taken up later along with HyTra framework - see comment from @chaubold (https://github.com/ilastik/ilastik/pull/1329)

### DIFF
--- a/volumina/imageView2D.py
+++ b/volumina/imageView2D.py
@@ -19,6 +19,7 @@
 # This information is also available on the ilastik web site at:
 #		   http://ilastik.org/license/
 ###############################################################################
+from __future__ import division
 from PyQt4.QtCore import QPoint, QPointF, QTimer, pyqtSignal, Qt, QRectF
 from PyQt4.QtGui import QCursor, QGraphicsView, QPainter, QVBoxLayout, QApplication, QImage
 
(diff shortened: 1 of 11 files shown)


njase:
Tag: py2to3_v0.3
Details: Further fixes based on Based on numpy#7949 for PEP238 support
Test status:
unit tests passed/same as last commit on master

volumina/imageView2D.py
```diff
@@ -244,14 +245,14 @@ def _setdeaccelerateAxAy(self, x, y, a):
         y = abs(y)
         if x > y:
             if y > 0:
-                ax = int(x / y)
+                ax = x // y
                 if ax != 0:
```
stuarteberg: Just a small note: In this change, the new line is not equivalent to the old line because the return type changes from int to float.  As it turns out, it doesn't matter in this case because the return value is eventually converted to float anyway.  But I had to go hunting for that to verify that it wasn't a problem.

For this exercise, when the original code achieved integer division via an explicit cast, let's just leave it alone.  It's not wrong, even if it might be a little verbose in some cases.

njase: Thanks. I'll remember this

PR #171, opened 2016-09-23
facilitates making cropping markers editable or not

ilastikdev:
needed in ilastik/learning for making cropping markers non-editable in tracking training

### DIFF
--- a/volumina/imageScene2D.py
+++ b/volumina/imageScene2D.py
@@ -229,6 +229,7 @@ class ImageScene2D(QGraphicsScene):
         self.scene2data, isInvertible = self.data2scene.inverted()
         self._setSceneRect()
         self._tiling.data2scene = self.data2scene
+        self._tileProvider._onSizeChanged()
         QGraphicsScene.invalidate(self, self.sceneRect())
 
     @property


volumina/imageView2D.py
```diff
@@ -89,7 +89,7 @@ def hud(self, hud):
 
         scene.axesChanged.connect(hud.setAxes)
 
-    def __init__(self, parent, cropModel, imagescene2d):
+    def __init__(self, parent, cropModel, imagescene2d, editable):
```
akreshuk: I'd prefer that the editable flag is set later, rather than changing the signature of a generic method only to pass that to a crop model

volumina/volumeEditor.py
```diff
@@ -142,7 +142,7 @@ def lastImageViewFocus(self, axis):
         self._lastImageViewFocus = axis
         self.newImageView2DFocus.emit()
 
-    def __init__( self, layerStackModel, parent, labelsink=None, crosshair=True, syncAlongAxes=(0,1)):
+    def __init__( self, layerStackModel, parent, labelsink=None, crosshair=True, syncAlongAxes=(0,1), editable=True):
```
akreshuk: Same as above, I'd rather have a method like "setEditable(flag)" than change a generic method signature for just one applet use case.

PR #172, opened 2016-10-12
adds editability on/off for crop lines

ilastikdev:
needed in learning for training applet where editability must be turned off

### DIFF
--- a/volumina/croppingMarkers.py
+++ b/volumina/croppingMarkers.py
@@ -34,6 +34,7 @@ class CropExtentsModel( QObject ):
 
     def setEditable (self, flag):
         self._editable = flag
+        self.editableChanged.emit(flag)
 
     def __init__(self, parent):
         super( CropExtentsModel, self ).__init__( parent )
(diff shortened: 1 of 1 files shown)


stuarteberg:
Maybe this would be a good opportunity to remove the `editableChanged` signal, since it doesn't appear to do anything.

ilastikdev:
It is used to set editability and this may be done on a user toggle basis if I merge cropping and training applet (switching between training an dcropping modes), which is very likely, so I would leave it .

stuarteberg:
I was confused because the text `editableChanged.emit` does not appear anywhere in the volumina source code.  But now I see it is called externally, in `tracking/annotations/annotationsGui.py`.  The existence of a signal that is emitted by external code, rather than by its own class, is not very idiomatic.  Can that be changed?

ilastikdev:
done

stuarteberg:
OK, thanks.

I guess I still don't understand why it is preferable to keep a copy of `CropExtentsModel._editable` stored in `CroppingMarkers`.  In the latter class, replacing all uses of

```python
if self._editable:
```

with this:

```python
if self.crop_extents_model.editable:
```

...would eliminate bookkeeping logic like `CroppingMarkers.onEditableChanged()`, and so on.  I could be missing something, though.  Is there a particular reason to store the `editable` flag twice?

ilastikdev:
cropping markers is a parent of crop line so in the OO spirit I wanted to have self.editable instead of 
self.parent.crop_extents_model.editable

ilastikdev:
@stuarteberg 
can we merge this?

stuarteberg:
sure

PR #175, opened 2016-12-24
Py3base volumina

njase:
This pull request is only for the purpose of review


volumina/pixelpipeline/datasources.py
```diff
@@ -201,7 +201,7 @@ def wrapper(*args, **kwargs):
                 return func(*args, **kwargs)
             except Slot.SlotNotReadyError as ex:
                 # Translate lazyflow not-ready errors into the volumina equivalent.
-                raise IndeterminateRequestError, IndeterminateRequestError(ex), sys.exc_info()[2]
+                raise IndeterminateRequestError(IndeterminateRequestError(ex)).with_traceback(sys.exc_info()[2])
         wrapper.__wrapped__ = func # Emulate python 3 behavior of @functools.wraps
```
stuarteberg: This looks a little wacky...

volumina/pixelpipeline/datasources.py
```diff
@@ -353,7 +353,7 @@ def put( self, slicing, array ):
             inputKeys = [tag.key for tag in inputTags]
             transposedArray = taggedArray.withAxes(*inputKeys)
     
-            taggedSlicing = dict(zip('txyzc', slicing))
+            taggedSlicing = dict(list(zip('txyzc', slicing)))
```
stuarteberg: extra `list` probably not needed here... right?

volumina/utility/signalingDefaultDict.py
```diff
@@ -29,13 +29,13 @@ def __getitem__(self, key):        return self._dict[key]
...
-    def iterkeys(self):                return self._dict.iterkeys()
-    def itervalues(self):              return self._dict.itervalues()
-    def values(self):                  return self._dict.values()
+    def viewkeys(self):                return self._dict.keys()
+    def items(self):                   return list(self._dict.items())
+    def iteritems(self):               return iter(self._dict.items())
+    def keys(self):                    return list(self._dict.keys())
+    def iterkeys(self):                return iter(self._dict.keys())
+    def itervalues(self):              return iter(self._dict.values())
+    def values(self):                  return list(self._dict.values())
```
stuarteberg: The idea with this class is to mimic the `dict` API, so `keys()`, `values()`, `items()` should return iterators.  And we should delete `iterkeys()`, etc.

volumina/pixelpipeline/datasources.py
njase: This is an automatic change by python utility - 2to3

volumina/pixelpipeline/datasources.py
njase: This is an automatic change by python utility - 2to3
In principle I agree with you

volumina/utility/signalingDefaultDict.py
njase: This is again an automatic change by python utility - 2to3.
The change looks necessary to me from the principles of py3 stated in PEP469.
 See http://legacy.python.org/dev/peps/pep-0469/#id8 which says:
"In Python 3, the objects returned by d.keys(), d.values() and d. items() provide a live view of the current state of the underlying object, rather than taking a full snapshot of the current state as they did in Python 2. This change is safe in many circumstances, but does mean that, as with the direct iteration API, it is necessary to avoid adding or removing keys during iteration, in order to avoid encountering the following error..."

PR #177, opened 2017-02-01
Setting tiling block size based on slice shape

JaimeIvanCervantes:
When loading videos with high resolution (eg: 3072x3072 larvae videos), the default current tiling block size of 256 creates a lot of tiles and therefore scrolling from frame to frame becomes really slow. We are now choosing the block size based on slice shape (only if it exceeds the default value 256). I tested this of 2D and 3D data sets.

@stuarteberg I think the code is safe, but could you review it just in case?

### DIFF
--- a/volumina/imageScene2D.py
+++ b/volumina/imageScene2D.py
@@ -305,9 +305,15 @@ class ImageScene2D(QGraphicsScene):
         and when the underlying data changes.
 
         """
+        DEFAULT_TILE_WIDTH = 512 
+        
         self.resetAxes(finish=False)
+        
+        tileWidth = self.tileWidth
+        if self.tileWidth is None:
+            tileWidth = DEFAULT_TILE_WIDTH
 
-        self._tiling = Tiling(self._dataShape, self.data2scene, name=self.name)
+        self._tiling = Tiling(self._dataShape, self.data2scene, name=self.name, blockSize=tileWidth)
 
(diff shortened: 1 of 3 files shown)


stuarteberg:
I agree we need something better than hard-coding the tile shape as we do now, but I don't think this change is the answer.  What would happen when viewing a 10k x 10k EM dataset (or a 100k by 100k EM dataset...)?  In that case, this proposed solution would make the tile size really huge.

I think we need to give the user control over this.  For now, a config file approach (read on startup) would be fine, I think.

JaimeIvanCervantes:
I am now allowing the user to set the tile size from the `View>Set Tile Width...` menu.

I also set the default block-size to 512px.

stuarteberg:
Were you able to figure out why the "Show Tiling" menu item wasn't updating properly?  Can it be fixed?

JaimeIvanCervantes:
Yes, It wasn't working because I was working on top of the latest commit where I was choosing the tile block-size based on the resolution, and it was rewriting the tile width set by the user. I removed this commit and everything works fine now.

stuarteberg:
I tried your branch, and it doesn't seem to be working for me.

Also, we should add the menu item in a different position in the menu.  Currently, it's in the "Only for selected view" section, which isn't right because it will affect all views.  (Side note: We should consider adding it in both places, in case the user wants to change the tiling on one view but not the others....)

PR #179, opened 2017-03-16
pyqt signal for mouse release added to cropping markers - needed for …

ilastikdev:
…automatic update of crops and filtering of annotations

toolTip signal added to ClickReportingInterpreter

needed for learning with tracking


ilastikdev:
this facilitates tooltips inside an image
there was no tooltip mechanism before that worked on an image

ilastikdev:
tooltip mechanism changes moved to an independent PR

PR #180, opened 2017-03-16
reporting interpreter added

ilastikdev:
it responds to mouse clicks and tooltips


akreshuk:
I asked for this change for the previous pull request and I was wrong: the ClickReportingInterpreter is the right place to do it.

PR #182, opened 2017-05-31
Deregister shortcuts for time scrolling only if they were registered …

chaubold:
…before, fixes ilastik/ilastik#1527

I tested it by changing axes orders randomly between `xyzc`, `zyxc`, and `txyc`.

Follow ups:
* Specifying axes order `tzyc` without `x` also works, but leads to the crash denoted in ilastik/ilastik#1528
* When switching from a time to a spatial axes, _mouse wheel scrolling_ is disabled in that plane. See ilastik/ilastik#1529


akreshuk:
Followed the same steps as yesterday and no more errors.

PR #183, opened 2017-06-01
removing crops

ilastikdev:
merging this into master should be done when pgmlink is no longer needed on windows
(pgmlink needs crops and full annotations in each crop)

needed for 
https://github.com/ilastik/ilastik/pull/1531


chaubold:
Looks good to me.

PR #187, opened 2017-07-06
always update shape in volume editor

ilastikdev:
ensures proper shape change when, for example, changing from txy to zxy axes


stuarteberg:
Is it not possible to call the new `ensureMinimized()` function directly from within `dataShape.setter`?

(I think we may have discussed this yesterday, so apologies if I've already forgotten the answer.)

ilastikdev:
Yes, we did discuss it and the consensus was that this should be handled by the caller rather than the viewer (this also avoids minimizing/maximizing in cases other than axes renaming)

stuarteberg:
Is this ready for merge now?

PR #188, opened 2017-08-24
Do not handle mouse click events twice

chaubold:
Set click events to accepted when they are handled, because otherwise the right-click context
menu of object classification and tracking showed up twice.

I verified that this works for ObjectClassification and Tracking with ObjectClassification on Mac.
And it still accepts clicks between the HUD buttons.


k-dominik:
LGTM

tested it on my machine.

PR #191, opened 2017-09-21
mouse single left click taken as a double click

ilastikdev:
we need _default for left click also
and because there is nothing specific to the RightClick in onMousePressRight_default
we can use it for all mouse clicks
that are not handled with higher priority in the if statement

type of the mouse click is passed via event parameter


chaubold:
Looks good to me, the tracking and (now also) manual tracking work as expected.

Tests pass locally for me, the ones on Cirlce CI fail because it looks like we messed up our conda packages...

PR #192, opened 2017-10-12
fix DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead

FynnBe:


k-dominik:
Signature change (just for future reference...)

"""python
# Python 2
warnings.warn(message[, category[, stacklevel]])

# Python 3
 warnings.warn(message, category=None, stacklevel=1, source=None)
"""
Defaults are the same as far as I can tell.

k-dominik:
btw. do you see any convention for when to use `warnings.warn` vs `logger.warning`? `imagesources.py` has both...

k-dominik:
@FynnBe , sorry about my first comment. My  grepping was wrong. You have reduced the number of warnings during testing from 261 to 237 for me :) so yay

chaubold:
@k-dominik if you use these kind of PRs of all three repos, then there are no deprecation warnings of the `warn` method left.

But for the `warnings.warn` vs `logger.warning` I am not su sure...

PR #193, opened 2017-11-07
Pyqtgraph rebased

akreshuk:
Replace the 3d widget by a pyqtgraph-based one


chaubold:
should be ready to merge, but we might want to wait until we can finally see a mesh on windows in carving?

PR #194, opened 2017-11-10
White color background in 3D widget

k-dominik:
Added the potential to set the background color via the config, with white as default.


chaubold:
works fine here!

PR #198, opened 2017-12-20
Show 3D widget only when necessary

wolny:
- [x] Add checkbox for toggling the 3D widget
- [x] Enable 3D widget via flag passed to the `VolumeEditor`

See https://github.com/ilastik/ilastik/issues/1666 for more details

Depends on https://github.com/ilastik/ilastik/pull/1686

out

### DIFF
--- a/volumina/view3d/overview3d.py
+++ b/volumina/view3d/overview3d.py
@@ -31,13 +31,15 @@ class Overview3D(QWidget):
     reinitialized = pyqtSignal()  # TODO: this should not be necessary: remove
     dock_status_changed = pyqtSignal(bool)
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, is_3d_visible=False, *args, **kwargs):
         super(QWidget, self).__init__(*args, **kwargs)
         cls, _ = loadUiType(join(split(__file__)[0], "ui/view3d.ui"))
         self._ui = cls()
         self._ui.setupUi(self)
 
         self._view = self._ui.view
+        # Set the visibility of 3D widget
+        self._view.setVisible(is_3d_visible)
         self._progress = self._ui.progress
         self._progress.setVisible(False)  # this can't be set in QDesigner for some reason
         self._mesh_generator_thread = None  # the thread need to be stored so it doesn't get destroyed when out of scope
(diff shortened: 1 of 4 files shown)


k-dominik:
ye looking good. I guess it is a good idea to keep this PR open until you integrate those changes with Carving. I suspect there will be some slight changes, e.g. synchronize the checkbox status to the `is_3d_visible` argument of the `__init__`...

wolny:
@k-dominik I dug up the code from around Xmas which enables 3D widget only for carving workflow (we can easily add another workflow types as well). It involves some ilastik changes to pass around the necessary context information (i.e. workflow class), but that's the easiest way I found to do it without too much refactoring. Lemme know what you think ;)

wolny:
This part is good to go IMO. Could you have another look @k-dominik? We need to merge this first, so that the https://github.com/ilastik/ilastik/pull/1686 build could pass.

volumina/volumeEditor.py
```diff
@@ -139,7 +139,7 @@ def lastImageViewFocus(self, axis):
         self._lastImageViewFocus = axis
         self.newImageView2DFocus.emit()
 
-    def __init__( self, layerStackModel, parent, labelsink=None, crosshair=True, syncAlongAxes=(0,1)):
+    def __init__( self, layerStackModel, parent, labelsink=None, crosshair=True, add_3d_widget=False, syncAlongAxes=(0,1)):
```
k-dominik: Naming the parameter `add_3d_widget` is a bit unfortunate; a 3D view is added in any case, it's just not visible, if I understand the code correctly. Maybe sticking to `is_3d_visible` or even more explicit like `3d_widget_visible` would be better

wolny: totally agree!

volumina/view3d/overview3d.py
```diff
@@ -31,13 +31,17 @@ class Overview3D(QWidget):
     reinitialized = pyqtSignal()  # TODO: this should not be necessary: remove
     dock_status_changed = pyqtSignal(bool)
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, is_3d_visible=False, *args, **kwargs):
```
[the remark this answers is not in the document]
wolny: ✅

volumina/volumeEditor.py
wolny: ✅

k-dominik:
I've looked at the code and played around with the carving workflow a bit. Works as expected, cool!

wolny:
Thanks for the review. Cool that we fixed the 2D bug on the way 🙌

PR #205, opened 2018-05-29
make sure to manipulate slicingPos or cursorPos as list (not tuple)

FynnBe:
scrolling lead to TypeError: 'tuple' object does not support item assignment


FynnBe:
Hmm.. Unfortunately I cannot reproduce it. It definitely did happen though, and slicingPos and cursorPos are supposed to be lists. I merge to cover that elusive rare case where somewhere somehow they become tuples...

PR #215, opened 2018-12-11
Fix IO on closed stream (Windows)

m-novikov:
I am not sure why colorama is embeded into our codebase, maybe it's better to migrate to upstream version in the future


k-dominik:
LGTM

PR #217, opened 2018-12-19
Fix getMainWindowBehaviour. Get the first biggest window

m-novikov:
ShortcutManager relies on function getMainWindow to detect to which window attach shortcut listener.
So in our case, we had two non QMainWindow widgets with the same size which lead to a crash.


m-novikov:
Closed in favor of #218

PR #218, opened 2018-12-20
Pytest migration

m-novikov:


m-novikov:
So from what I understood travis CI runs Windows Server Core images. As a result GUI isn't available.
Probably should stick with appveyor for build for the time being.

k-dominik:
Cool, could you remove windows from travis then and add a config for appveyor?

m-novikov:
> the perks of having a tiling window manager. Just out of curiosity - what will happen to
> 
> [volumina/volumina/utility/shortcutManager.py](https://github.com/ilastik/volumina/blob/c500bc07f9dde45aad2d54464738e5af1cbfa776/volumina/utility/shortcutManager.py#L181-L184)
> 
> Lines 181 to 184 in [c500bc0](/ilastik/volumina/commit/c500bc07f9dde45aad2d54464738e5af1cbfa776)
> when `getMainWindow` returns `None`

It will silently do nothing.
Forced volumina to raise exception if `getMainWindow` fails to produce result.
We could return None in this case, but it better to fail it shouldn't affect users, but for developers it would be easier to unserstand what's happening without chasing `None`.

k-dominik:
@m-novikov: Any idea on how to get `xvfb` to run?

PR #222, opened 2019-03-10
Ensure visible edges are clickable

stuarteberg:
The edge labeling UI (e.g. in the Multicut workflow) has always had an annoying problem, which is that sometimes you can't seem to click on an edge label, even when you've zoomed in on it all the way.  Somehow nearby edges respond to the click instead of the edge you intended.

Although the edge is drawn with a "cosmetic" pen, it has a a non-zero (but invisible) width, to permit users to click the edge without hitting the target perfectly.  The problem is that this invisible portion might overlap with the invisible portion of nearby edges, and whichever one happens to be "on top" in the Z-order is the one that responds to the click.

This PR implements a simple hack to make the UI more usable: If an edge is clicked "off" (invisible), it is sent to the bottom of the Z-order.  Even if it wasn't the edge you meant to click, at least the one you intended to click is now accessible.  (You still might accidentally turn off an edge you didn't intend, but hopefully it's still clickable at a different location.  For really tiny edges, surrounded by larger edges, this hack makes most of them clickable *somewhere*.)

If this solution doesn't suffice, it might be helpful to reduce the [width of the pen][1] that's used to draw the edges.  (Of course, that will make all edges tricky to click.)  Other suggestions welcome.

[1]: https://github.com/ilastik/ilastik/blob/0183986d560992ed99650b13a600e5e629229ac9/ilastik/applets/edgeTraining/edgeTrainingGui.py#L207

While making this change, I discovered a bug that could cause many invisible edges to be "redrawn" when they didn't need it, leading to bad performance.  With that fixed, the labeling UI is a little smoother than it was before.


volumina/pixelpipeline/imagesources.py
```diff
@@ -689,30 +689,32 @@ def request( self, qrect, along_through=None ):
...
     def __init__( self, layer, arraySource2D ):
-        from volumina.layer import SegmentationEdgesLayer
+        from volumina.layer import SegmentationEdgesLayer, LabelableSegmentationEdgesLayer
         assert isinstance(layer, SegmentationEdgesLayer)
         
         super( SegmentationEdgesItemSource, self ).__init__(layer.name)
         self._arraySource2D = arraySource2D
         self._arraySource2D.isDirty.connect(self.setDirty)
         self._layer = layer
+        self._is_clickable = isinstance(self._layer, LabelableSegmentationEdgesLayer)
```
m-novikov: It's better to avoid `isinstance` checks, as they can be confusing because they provide implicit interface.
We can define it explicitly for example with property `Layer.has_clickable_item`
Also, this information can be omitted here, because we pass `self._layer` as an argument to `SegmentationEdgesItemRequest`

volumina/pixelpipeline/imagesources.py
```diff
@@ -689,30 +689,32 @@ def request( self, qrect, along_through=None ):
...
     def image_type(self):
         return SegmentationEdgesItem
 
 class SegmentationEdgesItemRequest(object):
-    def __init__(self, arrayreq, layer, rect):
+    def __init__(self, arrayreq, layer, rect, is_clickable):
         self.rect = rect
         self._arrayreq = arrayreq
         self._layer = layer
+        self._is_clickable = is_clickable
```
m-novikov: Here we can instead define and drop everything above if we define an interface (Layer.has_clickable_items) explicitly.
```python
@property
def _is_clickable(self):
      return self._layer.has_clickable_items
```

volumina/utility/signalingDefaultDict.py
```diff
@@ -38,6 +36,18 @@ def iterkeys(self):                return iter(self._dict.keys())
...
 
+    def __getitem__(self, key):
+        try:
+            return self._dict[key]
+        except KeyError:
+            return self.default_value
+
+    def get(self, key, default=None):
+        try:
+            return self._dict[key]
```
m-novikov:
```python
return self._dict.get(key, default)
```

volumina/utility/segmentationEdgesItem.py
```diff
@@ -66,7 +71,33 @@ def handle_mouse_drag(self, id_pair, event):
...
         for id_pair in updated_path_ids:
-            self.path_items[id_pair].setPen( self.edge_pen_table[id_pair] )
+            item = self.path_items[id_pair]
+            pen = self.edge_pen_table[id_pair]
+            item.setPen( pen )
+            
+            if not self.is_clickable:
+                continue
+            
+            colliding = list(filter(lambda c: c.parentItem() is item.parentItem(), item.collidingItems()))
```
m-novikov:
```python
colliding = [c for c in item.collidingItems() if c.parentItem() is item.parentItem()]
```

volumina/utility/signalingDefaultDict.py
```diff
@@ -38,6 +36,18 @@ def iterkeys(self):                return iter(self._dict.keys())
     def itervalues(self):              return iter(self._dict.values())
     def values(self):                  return list(self._dict.values())
 
+    def __getitem__(self, key):
+        try:
+            return self._dict[key]
```
m-novikov:
```python
return self._dict.get(key, self.default_value)
```

volumina/utility/signalingDefaultDict.py
```diff
@@ -20,16 +19,15 @@ class SignalingDefaultDict( QObject ):
...
     updated = pyqtSignal(object) # set(updated_keys) 
     
-    def __init__(self, parent, default_factory, *args, **kwargs):
-        assert default_factory is not None, "You must provide a default_factory."
+    def __init__(self, parent, default_value):
         super( SignalingDefaultDict, self ).__init__(parent)
-        self._dict = defaultdict(default_factory, *args, **kwargs)
+        self._dict = {}
+        self.default_value = default_value
+        assert not callable(default_value), "Pass a default_value, not a default_factory"
```
m-novikov: We should leave it with callable.
Currently, factory function is used incorrectly it always returns the same object of `QPen`.
Which may lead to accidental changes to the default value,  it's fixable with callable, but not with a default value.
```python
p = QPen()
a = SignalingDefaultDict(None, p)
p1 = a['mypen']
p2 = a['myanotherpen']
p1.setColor(Qt.Red) # will change p2
```
also `defaultdict` will materialize requested keys, not sure how important this is

stuarteberg:
@m-novikov, thanks for the review!

You're right about `SignalingDefaultDict` -- it's confusing to call something a "default dict" if it doesn't actually behave like the standard `defaultdict` class.  In light of that, I've changed it to `SignalingDict`, and require the calling object to supply the default (via `.get()`).

In `SegmentationEdgesItem`, I want to avoid re-creating the default value (i.e. `default_factory()`) every time a value is missing because the default value will be needed quite often, and creating a new `QPen` for every (invisible) edge in the view would hinder responsiveness in the UI.

Regarding your note about `is_clickable`: In volumina, we already use the `Layer` subclass to determine how to construct the corresponding `ImageSource`, but in this case I was doing it in the wrong place.  Your comment helped me realize that I should do this in the more "standard" place -- in `imagesourcefactories.py`, where "multimethods" are used to map from layer types to image sources.  (I don't love this mechanism in general, but it's what volumina uses, so at least I'm now using the same mechanism as all the other layer types.)

m-novikov:
Great, thanks. :)
I have a couple of questions outside of scope of this pull request.
Is it possible to avoid drawing transparent paths alltogether excluding them from `generate_path_items_for_labels` result?
Why do we use `imagesourcefactories.py`, instead of creating/overriding a `Layer.createImageSource` method? Is there a cyclic dependency?

stuarteberg:
>Is it possible to avoid drawing transparent paths alltogether excluding them from `generate_path_items_for_labels` result?

We still need unlabeled edges to be clickable (in case the user wants to label them), and I think the only way to do that is to draw them with an invisible pen.

>Why do we use `imagesourcefactories.py`, instead of creating/overriding a `Layer.createImageSource` method? Is there a cyclic dependency?

That factory/multimethod mechanism dates back to the early days of volumina (before I joined the project).  Maybe the idea was that, this way, the viewer's "back end" (the pixelpipeline) might be more easily separated from its "front end" (the layer properties and its data sources), if we someday reimplement the way QImages are created and rendered onto the screen.  But if anyone ever does such refactor of the pixelpipeline code and decides to change the design, I won't complain.  Whoever does the refactor can judge what is clearest.

m-novikov:
There is side effect, if you click on same pixel it will produce different results cycling through "clickable edges" in this region.

PR #227, opened 2019-04-01
Remove colorama

m-novikov:


volumina/pixelpipeline/log.py
```diff
@@ -0,0 +1,3 @@
+import logging
+
+pixelpipeline_logger = logging.getLogger('volumina.pixelpipeline')
```
k-dominik: is this better then just using `logger = logging.getLogger(__file__)` everywhere?

volumina/pixelpipeline/imagesources.py
```diff
@@ -52,6 +46,28 @@
...
 
+
+logger = logging.getLogger(__name__)
+
+
+def log_request(func):
+    @functools.wraps(func)
+    def _log_request(source: "ImageSource", qrect, *args, **kwargs):
+        if CONFIG.verbose_pixelpipeline:
+            logger.error(
```
k-dominik: just out of curiosity, why do you use the error loglevel here?

m-novikov: Initially, it was in red (and most of the output prettifiers do stderr in red) and I want to avoid setting INFO log level for the logger, because it may log something else, except pipeline requests.

PR #228, opened 2019-04-01
Remove multimethods

m-novikov:


tests/pixelpipeline/test_datasourcefactories.py
```diff
@@ -0,0 +1,82 @@
...
+def test_lazyflow_factory(lazyflow_op, vigra, ndims):
+    shape = (10,) * ndims
+
+    array = rand(*shape).view(vigra.VigraArray)
+    array.axistags = vigra.defaultAxistags("txyzc"[:ndims])
+    lazyflow_op.inputs["Input"].setValue(array)
+
+    src = factories.createDataSource(lazyflow_op.Input)
+    assert isinstance(src, ds.LazyflowSource)
+    assert np.squeeze(np.ndarray(src._op5.Output.meta.shape)).shape == array.shape
```
k-dominik: hmmm you're just checking the shapes, which will only contain `10`s, so actually you could just compare the length. It would however be a stronger comparison, if shapes in the respective axes differ, so one can see that order is preserved, too.
Also while one is at it, one could check that the data is not altered, too (`numpy.testing.assert_array_equal...`)

m-novikov:
To be fair part of the test file is a copy of removed `datasourcefactories_test.py`.
Updated lazyflow test to reflect that it tests shape operator shape

tests/test_pixelpipeline/test_datasourcefactories.py
```diff
@@ -38,21 +23,40 @@ def vigra():
...
+DIMS = {
+    't': 2,
+    'c': 3,
+    'z': 10,
+    'x': 32,
+    'y': 64,
+}
+
+
+@pytest.mark.parametrize("dims,expected_shape", [
```
emilmelnikov: Not really an issue, but the following style could be somewhat easier to read:

```python
from types import SimpleNamespace

DIMS = SimpleNamespace(t=2, c=3, ...)

@pytest.mark.parametrize("dims,expected_shape", [
    [tuple("xy"), (1, DIMS.x, DIMS.y, 1, 1)],
    ....,
]
```
```

PR #235, opened 2019-05-09
Fixes "Image Export Options" export metadata preview

Tomaz-Vieira:
Creates OutputSlotMetaInfoDisplayWidget which properly handles preview of
export metadata.

Closes ilastik/ilastik/issues/1894


volumina/widgets/slotMetaInfoDisplayWidget.py
```diff
@@ -45,18 +45,27 @@ def initSlot(self, slot):
         self._refresh()
 
     def _refresh(self, *args):
+        shape = axes = dtype = ""
         if self._slot.ready():
             shape = tuple(self._slot.meta.getOriginalShape())
```
k-dominik: no biggy, really and I know there are no types in python, but since you initialized shape as a string, i would expect it to stay one?

Tomaz-Vieira: I agree, switching types is for savages D=
I should stop rushing these bug fixes

m-novikov:
Requesting some tests for this :)
See [example](https://github.com/ilastik/ilastik/blob/master/tests/widgets/test_appletDrawerToolBox.py) for `pytest-qt` this

tests/test_SlotMetaInfoDisplayWidget.py
```diff
@@ -0,0 +1,58 @@
...
+import vigra
+
+from lazyflow.graph import Graph
+from lazyflow.operators.operators import OpArrayPiper
+from lazyflow.operators.opReorderAxes import OpReorderAxes
+from volumina.widgets.slotMetaInfoDisplayWidget import (SlotMetaInfoDisplayWidget,
+                                                        OutputSlotMetaInfoDisplayWidget)
+
+def test_SlotMetaInfoDisplayWidget_shows_correct_info(qtbot):
+    shape = {'x':10, 'y':20, 'z':30, 'c':3}
```
k-dominik: i think in general there is the convention to use a space after a colon `:`

tests/test_SlotMetaInfoDisplayWidget.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
k-dominik: Our naming convention in volumina seems to be the other way round so, please rename to `slotMetaInfoDisplayWidget_test.py`

emilmelnikov:
Maybe rename `refresh` to something more indicative, like `update_labels`? Also, we could somewhat simplify the implementation:

```python
class SlotMetaInfoDisplayWidget(QWidget):
     ...

    def initSlot(self, ...):
        ...
        self._update_labels()

    def _update_labels(self):
        labels = dict(_shape=self.shapeDisplay, _axisOrder=self.axisOrderDisplay, _dtype=self.dtypeDisplay)
        if any(map(sip.isdeleted, self._labels.values())):
            return
        for name, label in labels.items():
            value = ""
            if self._slot.ready():
                value = str(getattr(self, name))
            label.setText(value)

    _shape = property(fget=lambda self: tuple(self._slot.meta.getOriginalShape()))
    _axisOrder = property(fget=lambda self: "".join(self._slot.meta.getOriginalAxisKeys()))
    _dtype = property(fget=lambda self: self._slot.meta.dtype.__name__)


class OutputSlotMetaInfoDisplayWidget(SlotMetaInfoDisplayWidget):
    _shape = property(fget=lambda self: "".join(self._slot.meta.getAxisKeys()))
    _axisOrder = property(fget=lambda self: "".join(self._slot.meta.getAxisKeys()))
```

k-dominik:
if you could now please run black against your changes, CI might be happy, too

PR #237, opened 2019-05-13
Auto deploy to gh pages

k-dominik:
this deploys the docs to gh-pages when merged on master.

Example: https://travis-ci.org/ilastik/volumina/jobs/531678308

there I had everything configure to build when on the branch "test"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,19 @@
 language: generic
-os:
-  - linux
-  - osx
+
 env:
   global:
     - CONDA_ROOT=$HOME/miniconda
     - TEST_ENV=test-env
     - TEST_ENV_PREFIX=$CONDA_ROOT/envs/$TEST_ENV
-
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -56,3 +68,49 @@ script:
...
+          fi
+        - echo ". $CONDA_ROOT/etc/profile.d/conda.sh" >> ~/.bashrc
+        - source ~/.bashrc
+        - conda config --set always_yes yes --set changeps1 no
+        - conda update -q conda
+        - |
+          if [[ ! -d deploy ]]; then
+            conda create -q -n deploy -c ilastik-forge -c conda-forge ilastik-dependencies-no-solvers sphinx;
+          else
+            conda install -y -n deploy -c ilastik-forge -c conda-forge ilastik-dependencies-no-solvers sphinx;
```
m-novikov: Do we really need to install ilastik deps in this case? Can we install only sphinx?

k-dominik: you spot every bit of lazyness I have in me. It didn't work without, so my quick fix was to download the whole thing. There are parts in `docs/conf.py` that require `lazyflow` for nicer display of string names or something... I'll investigate

PR #239, opened 2019-06-18
added MockGLView to replace GLView if OpenGL < 2.0

k-dominik:
ilastik carving gui tests on Windows CI (appveyor) failed because
the 3D widget would raise an error when drawing 3D stuff, as the
provided GL version (1.0, 1.1) is not enough.

* Added a check to GLView, to get the current OpenGL version
* Added a MockGLView that is replaced in Overview3D in case version
  is not sufficient
* Config/environment variables:
  * Config values can now be overriden via environment variables
  * `VOLUMINA_VERBOSE_PIXELPIPELINE=1`, `VOLUMINA_SHOW_3D_WIDGET=0`

Related to  https://github.com/ilastik/ilastik/issues/1297 https://github.com/ilastik/ilastik/issues/1885

Todo:

- [x] add test :)

### DIFF
--- a/volumina/config.py
+++ b/volumina/config.py
@@ -28,6 +28,9 @@ import os
 default_config = """
 [pixelpipeline]
 verbose: false
+
+[viewport]
+use_opengl: true
 """
 
 _cfg = configparser.ConfigParser()
@@ -40,10 +43,25 @@ if os.path.exists(userConfig):
 class Config:
     def __init__(self, cfg):
         self._cfg = cfg
+        self._env = os.environ
 
     @property
     def verbose_pixelpipeline(self):
-        return self._cfg.getboolean("pixelpipeline", "verbose")
+        return self._get_boolean("pixelpipeline", "verbose")
+
+    @property
+    def use_opengl_viewport(self):
+        return self._get_boolean("viewport", "use_opengl")
+
+    def _get_boolean(self, section: str, option: str) -> bool:
+        val = self._env.get(f"{section.upper()}_{option.upper()}")
+        if val is not None:
+            try:
+                return bool(int(val))
+            except ValueError:
+                pass
+
+        return self._cfg.getboolean(section, option)
 
 
 CONFIG = Config(_cfg)
--- a/volumina/config.py
+++ b/volumina/config.py
@@ -53,13 +53,13 @@ class Config:
 
     def _get_boolean(self, section: str, option: str) -> bool:
         val = self._env.get(f"{section.upper()}_{option.upper()}")
-        if val is not None:
-            try:
-                return bool(int(val))
-            except ValueError:
-                pass
+        if val is None:
+            return self._cfg.getboolean(section, option)
 
-        return self._cfg.getboolean(section, option)
+        try:
+            return bool(int(val))
+        except ValueError as e:
+            raise ValueError(f"environment variable {val!r} is not a boolean") from e
 
 
 CONFIG = Config(_cfg)
--- a/volumina/config.py
+++ b/volumina/config.py
@@ -26,11 +26,9 @@ import configparser
 import os
 
 default_config = """
-[pixelpipeline]
-verbose: false
-
-[viewport]
-use_opengl: true
+[volumina]
+pixelpipeline_verbose: false
+show_3d_widget: true
 """
 
 _cfg = configparser.ConfigParser()
@@ -47,11 +45,11 @@ class Config:
 
     @property
     def verbose_pixelpipeline(self):
-        return self._get_boolean("pixelpipeline", "verbose")
+        return self._get_boolean("volumina", "pixelpipeline_verbose")
 
     @property
-    def use_opengl_viewport(self):
-        return self._get_boolean("viewport", "use_opengl")
+    def show_3d_widget(self):
+        return self._get_boolean("volumina", "show_3d_widget")
 
     def _get_boolean(self, section: str, option: str) -> bool:
         val = self._env.get(f"{section.upper()}_{option.upper()}")
--- a/volumina/view3d/glview.py
+++ b/volumina/view3d/glview.py
@@ -2,13 +2,19 @@ from pyqtgraph.opengl import GLViewWidget
 
 from PyQt5.QtCore import pyqtSignal
 from PyQt5.QtGui import QVector4D
+from PyQt5.QtWidgets import QLabel
 
 from volumina.view3d.slicingplanes import SlicingPlanes
 from volumina.view3d.axessymbols import AxesSymbols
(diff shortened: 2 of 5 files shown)


volumina/view3d/glview.py
```diff
@@ -39,6 +47,31 @@ def __init__(self, parent=None):
         size_policy.setRetainSizeWhenHidden(True)
         self.setSizePolicy(size_policy)
 
+    def gl_version_sufficient(self) -> Tuple[bool, str]:
```
emilmelnikov: It seems like [`glGetString(GL_VERSION)`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetString.xhtml) can also a return version string in the `MAJOR.MINOR.RELEASE` form.
Also, comparing floats works in this case, but feels somewhat hacky.

Maybe do it in the following way?

```python
_MININUM_GL_VERSION = (2, 0, 0)

def gl_version(self):
    version = tuple(map(int, glGetString(GL_VERSION).decode().partition(" ")[0].split(".")))
    assert len(version) in (2, 3)
    if len(version) == 2:
        version = version + (0,)
    return version
```

k-dominik: True that, but I didn't find any guarantees in the openGL docs. But I'll add something in those terms.

volumina/view3d/overview3d.py
```diff
@@ -62,6 +76,15 @@ def _adjust_axes(x, y, z):
         """
         return x, y, z
 
+    def _remove_3d_view(self, display_string: str):
```
Tomaz-Vieira: I wonder if, rather than allowing a normal GlView to be created, and then have it replaced here, if it wouldn't be better to hide the declaration of the real GLView in the cases where OpenGL is not supported.
```python
if gl_version_sufficient:
    class GlView:
        "Real GlView"
else:
    class GlView:
        "Fake GlView"
```

That way other consumers of GLView will also have their implementation mocked automatically, plus we might dodge some traps of not setting up the mock properly, or failing to tear down the real one the correct way

emilmelnikov: Can we create both and use only one of them?

```python
class GLView_Real:
    ...
class GLView_Fake:
    ....
GLView = GLView_Real if OPENGL_VERSION >= OPENGL_VERSION_MIN else GLView_Fake
```

k-dominik: thought about it at the beginning, too, but found the "big if" pattern very ugly (just aesthetically), but I see the benefits. I'll give it a go, since you guys not seem to be repelled :)

k-dominik: Okay, I did not have success in checking GL version setting the context myself. Does anyone else want a go with it? Otherwise I'll merge this an open an issue.

emilmelnikov: Maybe we can just set a special environment variable `VOLUMINA_ENABLE_STUB_GLVIEW=1` and check for it in the code? It seems like this is needed only for tests, and if somebody wants to actually use volumina with an old OpenGL version, they need to install drivers anyway.

k-dominik: well, the other use case is x-forwarding which was reportedly being problematic.

Tomaz-Vieira: > Okay, I did not have success in checking GL version setting the context myself. Does anyone else want a go with it? Otherwise I'll merge this an open an issue.

Hell, I always wanted to suffer under OpenGL =D
If you open the issue I'll get to it as soon as I get some time to breathe. But it's still useful to get this merged anyway

k-dominik:
So, I made it configurable, either via `.voluminarc` or an environment variable `FEATURES_USE_OPENGL_VIEWPORT`.
Now that I read that name, it should probably not contain the term `FEATURES`.
I removed all the version checking code. Need to add a test...

k-dominik:
okay, I force-pushed this again, since it became to convoluted. @Tomaz-Vieira, @emilmelnikov if you'd kindly take a look again :)

volumina/config.py
```diff
@@ -40,10 +43,25 @@
...
     def verbose_pixelpipeline(self):
-        return self._cfg.getboolean("pixelpipeline", "verbose")
+        return self._get_boolean("pixelpipeline", "verbose")
+
+    @property
+    def use_opengl_viewport(self):
+        return self._get_boolean("viewport", "use_opengl")
+
+    def _get_boolean(self, section: str, option: str) -> bool:
+        val = self._env.get(f"{section.upper()}_{option.upper()}")
```
emilmelnikov:
```python
val = self._env.get(f"{section.upper()}_{option.upper()}")
if val is None:
    return self._cfg.getboolean(section, option)
try:
    return bool(int(val))
except ValueError as e:
    raise ValueError(f"environment variable {val!r} is not a boolean") from e
```

volumina/view3d/glview.py
```diff
@@ -190,3 +196,46 @@ def mouseReleaseEvent(self, event):
...
+    def set_shape(self, shape):
+        pass
+
+    shape = property(fset=set_shape)
+
+    def toggle_slice(self, axis, visible):
+        pass
+
+
+GLView = volumina.config.CONFIG.use_opengl_viewport and GLViewReal or GLViewMock
```
emilmelnikov: Boolean operators are somewhat confusing here. Maybe change it to `if` expression/statement?

volumina/view3d/glview.py
```diff
@@ -190,3 +196,46 @@ def mouseReleaseEvent(self, event):
...
         self._slice_planes.release()
         return GLViewWidget.mousePressEvent(self, event)
+
+
+class GLViewMock(QLabel):
+    slice_changed = pyqtSignal()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.setText("3D widget disabled via $HOME/.voluminarc or environment variable\n" "FEATURES_USE_OPENGL_WIDGET.")
```
emilmelnikov:
```suggestion
        self.setText("3D widget disabled via $HOME/.voluminarc or environment variable" "<pre>FEATURES_USE_OPENGL_WIDGET</pre>")
```

volumina/config.py
k-dominik: thx! that's better

volumina/config.py
```diff
@@ -28,6 +28,9 @@
 default_config = """
 [pixelpipeline]
 verbose: false
+
+[viewport]
+use_opengl: true
```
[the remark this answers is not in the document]
k-dominik: went for "VOLUMINA_SHOW_3D_WIDGET", I think that's more readable than `VOLUMINA_STUB_GLVIEW".

PR #240, opened 2019-06-27
don't use 3d widget in ci

k-dominik:
Summary:

* use `devenv.py` to spin up test environment in appveyor and travis
* scratch circleci builds, linux is already covered by travis


m-novikov:
Apparently culprit is `conda remove ilastik-meta` should be  `conda remove --force ilastik-meta` to avoid deleting all deps of meta

k-dominik:
funny, I recall already changing that somewhere... good find :)

PR #242, opened 2019-07-08
fix properly use config to configure 3D view

k-dominik:
Related to #239 #240 

there was a bug where `volumina.config.Config` was directly called
and not instantiated. Apparently the `Config.show_3d_widget` always
returns a property-object that is always interpreted as `True`.

* Use proper `volumina.config.CONFIG` object (uppercase).
* Hide `_Config` class.

### DIFF
--- a/volumina/view3d/glview.py
+++ b/volumina/view3d/glview.py
@@ -204,7 +204,7 @@ class GLViewMock(QLabel):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.setText(
-            "3D widget disabled via $HOME/.voluminarc or environment variable" "<pre>FEATURES_USE_OPENGL_WIDGET</pre>"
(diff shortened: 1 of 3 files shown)


volumina/view3d/glview.py
```diff
@@ -204,7 +204,7 @@ class GLViewMock(QLabel):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.setText(
-            "3D widget disabled via $HOME/.voluminarc or environment variable" "<pre>FEATURES_USE_OPENGL_WIDGET</pre>"
```
m-novikov: should we also specify which of the `.voluminarc` options handles disabling the widget?

k-dominik: hm, since the default has it on, and this message will only be shown if it is off, I think it should be fine.
One the other hand, doesn't hurt to be more explicit.

PR #243, opened 2019-07-29
Replace multimethod with singledispatch

emilmelnikov:
Remove the `multimethod` decorator and use `functools.singledispatch`
instead: all current use cases of "multimethods" don't require multiple
dispatch.

Also, refactor `volumina.pixelpipeline.datasourcefactories`.


emilmelnikov:
Oops, it turns out that we already have #228 that does exactly the same thing but better, so I'm closing this.

PR #246, opened 2019-08-01
Fix duplicate mouse move events

emilmelnikov:
Mouse move events were delivered 2 times, and the second event was
delivered *after the next mouse move event*. For example, when dragging
the mouse to the right:

    mouseMoveEvent: x=1 y=42  # Event #1, first delivery
    mouseMoveEvent: x=3 y=42  # Event #2, first delivery
    mouseMoveEvent: x=1 y=42  # Event #1, second (duplicate) delivery
    ...

This behavior produced non-empty boxes even when clicking on the canvas
without dragging.

Related to https://github.com/ilastik/ilastik/pull/2066.

### DIFF
--- a/volumina/navigationController.py
+++ b/volumina/navigationController.py
@@ -19,21 +19,16 @@
 # This information is also available on the ilastik web site at:
 # 		   http://ilastik.org/license/
 ###############################################################################
-from __future__ import division
-from __future__ import absolute_import
-from builtins import range
-from PyQt5.QtCore import QObject, QTimer, QEvent, Qt, QPointF, pyqtSignal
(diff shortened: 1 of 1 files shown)


emilmelnikov:
> LGTM
> 
> I just want to note a _small side effect for multicut_ (which in my opinion is a _bugfix_, rather than a side effect). Previously it was possible to be in drag mode and all edges on the path would be marked according to the mouse button pressed. Now one really has to click. If we want this behaviour back, we should properly implement it and not rely on the interpreter pretending not to have handled the event.

Can we make an issue out of this?

PR #247, opened 2019-08-09
Draft cache source

m-novikov:


k-dominik:
nice! just checked out window/leveling on a 3D volume and this is smooth now

PR #248, opened 2019-08-20
Split tiling

m-novikov:


volumina/tiling/cache.py
```diff
@@ -0,0 +1,295 @@
...
+    def unsubsribe(self, fn):
+        try:
+            self._subsribers.remove(fn)
+        except ValueError:
+            pass
+
+
+class MultiCache:
+    """
+    A utility class for caching items in a of a dict-of-dicts
```
k-dominik:
```suggestion
    A utility class for caching items in a dict-of-dicts
```

volumina/tiling/cache.py
```diff
@@ -0,0 +1,295 @@
...
+    def __contains__(self, stack_id):
+        assert self._lock.locked(), "You must claim the _TileCache via a context manager before calling this function."
+        return stack_id in self._tileCache
+
+    def __len__(self):
+        assert self._lock.locked(), "You must claim the _TileCache via a context manager before calling this function."
+        return len(self._tileCache)
+
+    def tile(self, stack_id, tile_id):
+        assert self._lock.locked(), "You must claim the _TileCache via a context manager before calling this function."
```
k-dominik: I know this is not your code, but since you are touching it anyway, we might discuss whether asserts are a good idea here. `python -O` will remove those...

volumina/tiling/cache.py
```diff
@@ -0,0 +1,295 @@
...
+        self._caches = OrderedDict()
+        self._default_factory = default_factory
+        self.add(first_uid)
+
+    def add(self, uid) -> None:
+        if uid not in self._caches:
+            cache = defaultdict(self._default_factory)
+            self._caches[uid] = cache
+        else:
+            raise Exception("MultiCache.add: uid %s is already in use" % str(uid))
```
k-dominik: Okay, this has to go, raising exception is just not right! Every code fragment wanting to catch this would have to `except Exception....`

tests/test_tiling/test_cache.py
```diff
@@ -0,0 +1,111 @@
...
+        policy.set_size(2)
+        cache.add("mytestid1")
+        cache.add("mytestid2")
+        cache.add("mytestid3")
+
+        assert "mytestid1" not in cache
+        assert "mytestid2" in cache
+        assert "mytestid3" in cache
+
+    def test_displace_eviction(self, cache, policy):
```
k-dominik: duplicate test, please remove

tests/test_tiling/test_cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+        assert "mytestid2" not in cache
+        assert "mytestid1" in cache
+        assert "mytestid3" in cache
+
+    def test_iteration(self, cache):
+        cache.add("mytestid1")
+        cache.add("mytestid2")
+
+        keys = []
```
k-dominik:
```suggestion
        keys = list(cache)
```

volumina/tiling/cache.py
m-novikov: I will do follow up pull-request there are some libraries that provide syncornized decorator that do this.

volumina/tiling/cache.py
m-novikov: I will do a follow-up pull-request with this change. There are libraries that provide decorators for this task

k-dominik (approved):
LGTM, nice cleaning up/simplification! No side-effects as far as I can see and added guarantees through tests :+1:

volumina/tiling/cache.py
```diff
@@ -0,0 +1,299 @@
...
+        self._layerCacheDirty = MultiCache(default_factory=lambda: True, **kwargs)
+
+        # [stack_id][(ims, tile_id)] -> float
+        self._layerCacheTimestamp = MultiCache(default_factory=float, **kwargs)
+
+    @property
+    def maxstacks(self):
+        return self._maxstacks
+
+    def set_maxstacks(self, maxstacks):
```
emilmelnikov, line 149: Make this a setter property?

volumina/tiling/cache.py
```diff
@@ -0,0 +1,299 @@
...
+        layerCacheDirty: A cache of dirty bits for all layers in layerCache
+        layerCacheTimestamp: A cache of timestamps to track how recently each layer was needed.
+
+        tileCacheDirty: A cache of dirty bits for the composite tiles
+                        (i.e. for a given patch, if a single layer in the patch
+                        is dirty, then the tile for that patch is dirty)
+    """
+
+    def __init__(self, first_stack_id, sims, maxstacks):
+        self._lock = threading.Lock()
```
emilmelnikov, line 123: This probably will not make a difference, but I think it's safe to use a reentrant lock here.
```suggestion
        self._lock = threading.RLock()
```

volumina/tiling/cache.py
```diff
@@ -0,0 +1,299 @@
...
+    def __enter__(self):
+        self._lock.acquire()
+        return self
+
+    def __exit__(self, *args):
+        assert self._lock.locked()
+        self._lock.release()
+
+    def __contains__(self, stack_id):
+        assert self._lock.locked(), "You must claim the _TileCache via a context manager before calling this function."
```
emilmelnikov, line 161: I wonder if it makes sense to make such methods that require a lock "synchronized"?

```python
def __contains__(...):
    with self:
        return ...
```

tests/test_tiling/test_cache.py
```diff
@@ -0,0 +1,99 @@
...
+    @pytest.fixture
+    def cache(self, policy):
+        return MultiCache("testid", policy=policy)
+
+    def test_creation(self, policy):
+        cache = MultiCache("testid", policy=policy)
+
+    def test_uid_collision_raises(self, cache):
+        cache.add("mytestid")
+        with pytest.raises(Exception):
```
emilmelnikov, line 22: Can use more precise exception class here.

volumina/tiling/cache.py
```diff
@@ -0,0 +1,299 @@
...
+        self._subsribers.append(fn)
+
+    def unsubsribe(self, fn):
+        try:
+            self._subsribers.remove(fn)
+        except ValueError:
+            pass
+
+
+class MultiCacheException(Exception):
```
emilmelnikov, line 49: Rename the exception class to something more descriptive like `DuplicateKeyError`?

Can also move the error message to `__init__`:
```python
def __init__(self, key):
    super().__init__(f"Duplicate key {key}")
```

volumina/tiling/cache.py
```diff
@@ -0,0 +1,299 @@
...
+    pass
+
+
+class MultiCache:
```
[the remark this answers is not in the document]
m-novikov: `abc.Sequence` expects `__getitem__` method to be indexable, not keyable. Running tests fails with:
```    
def __getitem__(self, key):
>       return self._caches[key]
E       KeyError: 0
```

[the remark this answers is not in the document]
emilmelnikov: What about `Mapping` then?

volumina/tiling/cache.py
m-novikov: I prefer setter method, because Exceptions on setting an attribute usually unexpected behaviour, but in methods it's perfectly normal occurence. I would be conviused seeing
```pyton
try:
   obj.attr = 12
except ValueError:
   pass
```

volumina/tiling/cache.py
```diff
@@ -0,0 +1,299 @@
...
+    def __getitem__(self, key):
+        return self._caches[key]
+
+    def __iter__(self):
+        return iter(self._caches)
+
+    def __len__(self):
+        return len(self._caches)
+
+    def touch(self, uid):
```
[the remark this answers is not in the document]
m-novikov: This is the same behavior as [redis touch](https://redis.io/commands/touch) added docstring

volumina/tiling/cache.py
m-novikov: This change crashes volumina tests :(

volumina/tiling/cache.py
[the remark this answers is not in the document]
m-novikov: The question is what is the benefit of using `Mapping` interface here.
On the other hand, if I were to use `Mapping` it can be confusing because when I think about `Mapping` I expect a hashmap with certain behavior, which should not throw away keys.

volumina/tiling/cache.py
emilmelnikov: I think it's normal to have an exception when setting an attribute, considering Python's dynamic behaviour. For example, `a = numpy.zeros(4); a.shape = (2, 2)` works, but `a = numpy.zeros(4); a.shape = (2, 999)` raises `ValueError`. You can change an index in pandas dataframe in a similar fashion.

volumina/tiling/cache.py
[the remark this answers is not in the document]
emilmelnikov: `Mapping` does not guarantee hashmap-like behaviour, only existence of a certain methods. For example, cache objects in [cachetools](https://github.com/tkem/cachetools/) are mappings.

volumina/tiling/cache.py
```diff
@@ -39,35 +40,36 @@ def set_size(self, value):
...
 
 
 class MultiCache:
     """
     A utility class for caching items in a dict-of-dicts
     """
 
-    def __init__(self, first_uid, default_factory=lambda: None, policy=None):
+    def __init__(self, default_factory=lambda: None, policy=None):
+        if not policy:
```
emilmelnikov, line 59: Any reasons not to make `policy` a required argument?

volumina/tiling/cache.py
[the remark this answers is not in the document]
m-novikov: My point is mostly if I had a class where I used `Mapping` which can be substituted for either for `MultiCache` or some other `Mapping` I would then inherit from this abstract class. 
But the only place where we use `MultiCache` excepts `MultiCache` and not a `Mapping` so inheriting mapping doesn't really provide any benefit in terms of decoupling.

volumina/tiling/cache.py
m-novikov: I suppose this used as optimization, as RLock for some reason doesn't work with the current codebase it's basically a way to perform multiple cache operations within one "transaction"

volumina/tiling/cache.py
m-novikov: I think it's also a matter of style preference. Here setting property also could lead to evictions from the cache. I'll leave it as is.

PR #250, opened 2019-08-20
Split SourceABC to multiple interfaces

m-novikov:
This pull request splits SourceABС into multiple interfaces. Old SourceABC was used in Image, Slice and Data sources.
But they all had different function and different interfaces. To better communicate intent now there is
`DataSourceABC` for querying "raw" data
`Slice2DSourceABC` for slicing data sources and producing 2d arrays.
`ImageSourceABC` for converting 2d arrays into QImage


tests/test_utility/test_abc.py
```diff
@@ -0,0 +1,45 @@
...
+    def test_type_error_on_undefined_abstract_method(self):
+
+        with pytest.raises(TypeError):
+            self.IObject()
+
+    def test_abstract_signal_raises_type_error(self):
+        with pytest.raises(TypeError):
+            self.ISignal()
+
+    def test_abstract_signal_raises_type_error(self):
```
k-dominik: is this some pytest dark magic that makes multiple methods with the same name work? In any case, I'd prefer not to have that.

volumina/pixelpipeline/datasources/interface.py
```diff
@@ -0,0 +1,61 @@
...
+    because its underlying datasource is in an indeterminate state.
+    In such cases, the requester should simply ignore the error.
+    The datasource has the responsibility of sending a dirty notification
+    when the source is ready again.
+    """
+
+    pass
+
+
+class DataRequestABC(ABC):
```
k-dominik: Is there a good reason why this does not reside in `pixelpipeline.asyncabcs`?

tests/test_pixelpipeline/test_slicesources.py
```diff
@@ -0,0 +1,36 @@
...
+    def testRequest(self):
+        self.ss.setThrough(0, 1)
+        self.ss.setThrough(2, 2)
+        self.ss.setThrough(1, 127)
+
+        sl = self.ss.request((slice(None), slice(None))).wait()
+        self.assertTrue(np.all(sl == self.raw[1, :, :, 127, 2]))
+
+        sl_bounded = self.ss.request((slice(0, 3), slice(1, None))).wait()
+        self.assertTrue(np.all(sl_bounded == self.raw[1, 0:3, 1:, 127, 2]))
```
k-dominik:
```suggestion
        numpy.testing.assert_array_equal(sl_bounded, self.raw[1, 0:3, 1:, 127, 2])
```

volumina/pixelpipeline/slicesources.py
```diff
@@ -42,7 +43,7 @@
 logger = logging.getLogger(__name__)
 
 
-class SliceRequest(object):
+class SliceRequest(RequestABC):
```
k-dominik:
```suggestion
class PlanarSliceRequest(RequestABC):
```

volumina/pixelpipeline/imagepump.py
```diff
@@ -351,15 +351,15 @@ def __init__(self, layerStackModel, sliceProjection, sync_along=(0, 1, 2)):
 
     # mappings
     def layerToSliceSources(self, layer):
```
k-dominik: don't know if this should be renamed to be uber consistent

volumina/pixelpipeline/datasources/interface.py
m-novikov: Some requests (e.g. GrayscaleImageRequest) don't implement getResult method.

volumina/pixelpipeline/imagepump.py
m-novikov: renamed `layerToPlanarSliceSources`

volumina/pixelpipeline/datasources/interface.py
k-dominik: sure, I mean the whole interface - i'd say it qualifies as async?

m-novikov: I am not even sure that `asyncabc` is a reasonable name :) It should be also `interfaces.py` or something

m-novikov:
Moved interfaces of pixelpipeline to interface
Removed getResult method.
* results are already cached for every wait method
* calling getResult on not awaited request will return None, which is error-prone as we have temporal coupling here.

k-dominik:
> Removed getResult method.

great

PR #255, opened 2019-09-05
Naive cache for Lazyflow output slots

m-novikov:


volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+    def delete(self, key: str) -> None:
+        with self._lock:
+            self._cache.pop(key, None)
+
+    def keys(self):
+        return list(self._cache.keys())
+
+    def set(self, key, value) -> None:
+        with self._lock:
+            old_entry = self._cache.get(key)
```
emilmelnikov:
```suggestion
            old_entry = self._cache.get(key, self._MISSING)

            if old_entry is not self._MISSING:
```

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __len__(self):
+        return len(self._cache)
+
+    def touch(self, key: str) -> None:
+        with self._lock:
+            self._cache.move_to_end(key)
+
+    @classmethod
+    def register_type(cls, _type: Type[T]):
```
emilmelnikov: Is it really possible to guarantee that a value of a certain type has a definite size? I would assume that it's possible only if a type defines `__slots__`. Otherwise, anything could be attached to the value's `__dict__`.

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+        def _register(get_size_fn: Callable[[T], int]):
+            cls._cachable_types[_type] = get_size_fn
+
+        return _register
+
+    @property
+    def used_memory(self) -> int:
+        return self._mem
+
+    def clean(self) -> None:
```
emilmelnikov: `clean` is semantically very close to `clear`. Maybe rename this to `trim` or something similar?

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(self, mem_limit=_256_MB):
+        self._cache = OrderedDict()
+        self._mem_limit = mem_limit
+        self._mem = 0
+        self._lock = threading.RLock()
+
+    def get(self, key: str, default=None) -> Any:
+        with self._lock:
+            entry = self._cache.get(key, self._MISSING)
+            self._cache.move_to_end(key)
```
emilmelnikov:
```suggestion
            if entry is not self._MISSING:
                self._cache.move_to_end(key)
```

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @classmethod
+    def register_type(cls, _type: Type[T]):
+        def _register(get_size_fn: Callable[[T], int]):
+            cls._cachable_types[_type] = get_size_fn
+
+        return _register
+
+    @property
+    def used_memory(self) -> int:
```
emilmelnikov: Probably a good idea to expose `max_memory` as well.

volumina/pixelpipeline/datasources/cachesource.py
```diff
@@ -0,0 +1,114 @@
...
+    def numberOfChannels(self):
+        return self._source.numberOfChannels
+
+    def __repr__(self):
+        return f"<CachedSource({self._source})>"
+
+    def __eq__(self, other):
+        if other is None:
+            return False
+        return self._source is other._source
```
emilmelnikov: This should work with all objects:
```suggestion
        return self._source is getattr(other, "_source", None)
```

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+class KVCache(CacheABC):
+    _MISSING = object()
+    _cachable_types: Dict[T, Callable[[T], int]] = {}
+
+    class _Entry(NamedTuple):
+        obj: Any
+        size: int  # bytes
+
+    def __init__(self, mem_limit=_256_MB):
+        self._cache = OrderedDict()
```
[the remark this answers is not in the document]
m-novikov: Cache assumes ownership of the data stored, otherwise, it would be hard to extend for eviction cases.

volumina/utility/cache.py
m-novikov: This function should provide an estimator, I only plan to store ndarrays and QImages both have a way to get size in bytes +- some not relevant overhead.

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+from typing import NamedTuple, Dict, Any, Type, TypeVar, Callable
+
+import numpy as np
+
+
+T = TypeVar("T")
+_256_MB = 256 * 1024 * 1024
+
+
+class CacheABC(abc.ABC):
```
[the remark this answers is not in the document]
m-novikov: I touch method is a good idea, but I would like to avoid adding `__iter__` and ` __len__` as it's unnecessary for cache interface and we can get fun errors such as `RuntimeError: xxx size changed during iteration`

volumina/utility/cache.py
k-dominik: i'll throw in `truncate` ;)

volumina/utility/cache.py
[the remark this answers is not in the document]
k-dominik: could be circumvented with locks, probably

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+import sys
+import threading
+from collections import OrderedDict
+from typing import NamedTuple, Dict, Any, Type, TypeVar, Callable
+
+import numpy as np
+
+
+T = TypeVar("T")
+_256_MB = 256 * 1024 * 1024
```
k-dominik: should be configurable

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+        if entry is not self._MISSING:
+            return entry.obj
+        else:
+            return default
+
+    def delete(self, key: str) -> None:
+        with self._lock:
+            self._cache.pop(key, None)
+
+    def keys(self):
```
k-dominik:
```suggestion
    def keys(self) -> List:
```
with the appropriate import

volumina/utility/cache.py
```diff
@@ -0,0 +1,119 @@
...
+        return len(self._cache)
+
+    def touch(self, key: str) -> None:
+        with self._lock:
+            self._cache.move_to_end(key)
+
+    @classmethod
+    def register_type(cls, _type: Type[T]):
+        def _register(get_size_fn: Callable[[T], int]):
+            cls._cachable_types[_type] = get_size_fn
```
emilmelnikov:
```python
if _type in cls._cacheable_types:
    raise ValueError(f"type {_type.__class__.__name__} already registered")
```

volumina/utility/cache.py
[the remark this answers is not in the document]
emilmelnikov: Tt should be `Iterable[T]` then.

volumina/utility/cache.py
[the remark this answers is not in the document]
emilmelnikov: > otherwise, it would be hard to extend for eviction cases.

Why so? We can use finalizers to update the cache size.

volumina/utility/cache.py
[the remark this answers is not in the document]
k-dominik: shouldn't it be `List[T]` ?

volumina/utility/cache.py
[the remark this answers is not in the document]
emilmelnikov: Or just by copying the keyset. Also, iterating (and, therefore, getting the size) seems important: otherwise there is no way to find out the contents of a cache. Caches are just mappings anyway, so I would argue it's very Pythonic to make it behave like a dictionary.

volumina/pixelpipeline/datasources/cachesource.py
```diff
@@ -0,0 +1,114 @@
...
+        if not is_pure_slicing(slicing):
+            raise Exception("dirty region: slicing is not pure")
+        self.isDirty.emit(slicing)
+
+    @property
+    def numberOfChannels(self):
+        return self._source.numberOfChannels
+
+    def __repr__(self):
+        return f"<CachedSource({self._source})>"
```
emilmelnikov:
```suggestion
        return f"<CachedSource({self._source!r})>"
```

volumina/utility/cache.py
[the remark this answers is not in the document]
emilmelnikov: If we follow the `Mapping` interface, then it's `Iterable`.

emilmelnikov:
Maybe we can use [an external caching library](https://github.com/tkem/cachetools) instead of a custom implementation?

m-novikov:
Updated pull-request using cachetools LRUCache implementation.
Regarding locking problem, the lock should not only cover cache access but also creation of requests querying for new data, so that's why we can't use lock inside cache manager.

volumina/pixelpipeline/datasources/cachesource.py
```diff
@@ -0,0 +1,123 @@
...
+        self._source.isDirty.connect(self.isDirty)
+        self._source.numberOfChannelsChanged.connect(self.numberOfChannelsChanged)
+        self._source.isDirty.connect(self.clear)
+        self._source.numberOfChannelsChanged.connect(self.clear)
+
+    def clear(self, *args):
+        self._cache.clear()
+        self._req.clear()
+
+    def __cache_key(self, slicing):
```
emilmelnikov: We could use `(self._id, *slicing)` as a cache key, maybe with an additional check that `all(isinstance(s, collections.Hashable) for s in slicing)`. We also could use `id(self)` instead of `self._id = next(self.__counter)`.

volumina/pixelpipeline/datasources/cachesource.py
```diff
@@ -0,0 +1,123 @@
...
+            _, key_part = el.__reduce__()
+            parts.append(key_part)
+
+        return "::".join(str(p) for p in parts)
+
+    def request(self, slicing):
+        key = self.__cache_key(slicing)
+
+        with self._lock:
+            if key in self._cache:
```
emilmelnikov:
```python
result = self._cache.get(key)
if result is not None:
    return _CachedRequest(result)
return self._req.setdefault(key, _Request(self, slicing, key))
```

m-novikov: On `setdefault` line _Request __init__ would be evaluated. It bothers me that we as a result also create lazyflow Request. We don't submit it yet, but it seems brittle

volumina/pixelpipeline/datasources/cachesource.py
m-novikov: I don't see a reason to allow non string types as cache keys, in this case.

volumina/utility/cache.py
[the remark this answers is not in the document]
m-novikov: Regarding this, when we move object to Cache is assumes ownership of this object.
How would cache behave when we moved couple of slices up in the image, at this point we lost all references to arrays out of our field of view?

volumina/pixelpipeline/datasources/cachesource.py
```diff
@@ -0,0 +1,120 @@
...
+
+from PyQt5.QtCore import QObject, pyqtSignal
+
+from volumina.pixelpipeline.interface import DataSourceABC
+from volumina.slicingtools import is_pure_slicing
+from volumina.utility.cache import KVCache
+from volumina.config import CONFIG
+
+
+ARRAY_CACHE = KVCache(CONFIG.cache_size, getsizeof=sys.getsizeof)
```
k-dominik: are we sure we'll never get a view as data? in that case `sys.getsizeof` won't work.

```ipython
In [2]: x = numpy.ones((10000, 2000))                                                                                                      

In [3]: sys.getsizeof(x)                                                                                                                   
Out[3]: 160000112

In [4]: y = x.reshape((10000, 2000))                                                                                                       

In [5]: sys.getsizeof(y)                                                                                                                   
Out[5]: 112

In [6]: x.nbytes                                                                                                                           
Out[6]: 160000000

In [7]: y.nbytes                                                                                                                           
Out[7]: 160000000
```

or would you say that is not a problem?

The problem I see is that even if I delete `x`, `getsizeof(y)` will not return the true number

tests/test_pixelpipeline/test_datasources/test_cachesource.py
```diff
@@ -0,0 +1,66 @@
...
+        arr = np.arange(27).reshape(3, 3, 3)
+        source = DummySource(arr)
+        return mock.Mock(wraps=source)
+
+    @pytest.fixture
+    def cached_source(self, orig_source):
+        return CacheSource(orig_source)
+
+    def test_consecutive_requests_are_cached(self, cached_source, orig_source):
+        slicing = np.s_[1:2, 1:2, 1:2]
```
k-dominik: just a minor comment - I'm always weary of using symmetric arrays for testing, especially when only querying the center voxel.

k-dominik:
i'll trigger a rebuilt once the new `ilastik-dependencies-no-solvers` package is up

k-dominik:
just an observation:

* in thresholding, the final output layer is not automatically shown with this PR.

PR #256, opened 2019-09-24
Do not remove missing layers in LayerDelegate

emilmelnikov:
For some reason, `LayerDelegate.handleRemovedRows` could be called when
the model's data does not contain an item for the `EditRole`.

Also, `QAbstractItemModel.rowsAboutToBeRemoved` signal has an inclusive
right boundary, so parameters have been renamed to `first` and `last`,
which is consistent with the [Qt documentation](https://doc.qt.io/qt-5/qabstractitemmodel.html#rowsAboutToBeRemoved).

Closes https://github.com/ilastik/ilastik/issues/1997.

### DIFF
--- a/volumina/widgets/layerwidget.py
+++ b/volumina/widgets/layerwidget.py
@@ -376,12 +376,12 @@ class LayerDelegate(QStyledItemDelegate):
         else:
             QStyledItemDelegate.setModelData(self, editor, model, index)
 
-    def handleRemovedRows(self, parent, start, end):
-        for row in range(start, end):
-            itemData = self._listModel.itemData(self._listModel.index(row))
-            layer = itemData[Qt.EditRole]
-            del self._editors[layer]
-            assert isinstance(layer, Layer)
+    def handleRemovedRows(self, _parent, first, last):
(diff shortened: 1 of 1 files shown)


k-dominik:
> > good catch, how is this involved in fixing [ilastik/ilastik#1997](https://github.com/ilastik/ilastik/issues/1997) though (sorry for not seeing that)?
> 
> See [ilastik/ilastik#1997 (comment)](https://github.com/ilastik/ilastik/issues/1997#issuecomment-533579327).

maybe this should be a new issue then, I cannot reproduce it with the instructions given in the comment. Being consistent with the QT specs there is good enough :+1:

emilmelnikov:
> maybe this should be a new issue then, I cannot reproduce it with the instructions given in the comment. Being consistent with the QT specs there is good enough

Please try to reproduce using the step sequence from https://github.com/ilastik/ilastik/issues/1997#issuecomment-534503485. I was able to reproduce this issue twice in a row.

emilmelnikov:
Merging this PR because we have been able to reproduce the issue with @k-dominik.

PR #257, opened 2019-09-26
PreferencesManager: allows setting file path, cleans up and adds tests

Tomaz-Vieira:
Allows instances of `PreferencesManager` to have their preferences path specified. This is useful for running tests that are truly independent, that is, that don't use tour local preferences files, which might alter the behavior of local tests.

This fixes a timeout issue with `testPixelClassificationGui.py`, because by creating a new `PreferencesManager` ilastik will not remember the last internal dataset path used with some test cases, which will prompt it to show a popup that allows the user to pick an internal dataset. If that popup never shows up, some tests which are expecting it might timeout.

### DIFF
--- a/tests/test_PreferencesManager.py
+++ b/tests/test_PreferencesManager.py
@@ -1,0 +1,42 @@
+import pytest
+from pathlib import Path
+
+from volumina.utility.preferencesManager import PreferencesManager
+
+
+def test_same_path_returns_same_instance(tmp_path, monkeypatch):
+    preferences_file_path = tmp_path / "my_preferences"
+    preferences_file_path.touch()
+
+    manager_absolute_path = PreferencesManager(preferences_file_path.absolute())
+
+    monkeypatch.chdir(tmp_path)
+    manager_relative = PreferencesManager(preferences_file_path.name)
+
+    assert id(manager_absolute_path) == id(manager_relative)
+
+    preferences_file_path_2 = tmp_path / "my_preferences2"
+    preferences_file_path_2.touch()
+    manager_absolute_path_2 = PreferencesManager(preferences_file_path_2)
+    manager_relative_2 = PreferencesManager(preferences_file_path_2.name)
+
+    assert id(manager_absolute_path_2) == id(manager_relative_2)
+    assert id(manager_absolute_path) != id(manager_absolute_path_2)
+    assert id(manager_relative) != id(manager_relative_2)
+
+
+def test_default_preferences_writing_into_home_dir(monkeypatch, tmp_path_factory):
+    monkeypatch.setenv("HOME", tmp_path_factory.mktemp("some_temp_dir").as_posix())
+
+    prefsMgr = PreferencesManager()
+    with PreferencesManager() as prefsMgr:
+        prefsMgr.set("Group 1", "Setting1", [1, 2, 3])
+        prefsMgr.set("Group 1", "Setting2", ["a", "b", "c"])
+        prefsMgr.set("Group 2", "Setting1", "Forty-two")
+
+    prefsMgr2 = PreferencesManager()
+    assert id(prefsMgr) == id(prefsMgr2), "It's supposed to be a singleton!"
+
+    assert prefsMgr2.get("Group 1", "Setting1") == [1, 2, 3]
+    assert prefsMgr2.get("Group 1", "Setting2") == ["a", "b", "c"]
+    assert prefsMgr2.get("Group 2", "Setting1") == "Forty-two"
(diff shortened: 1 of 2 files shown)


m-novikov:
I think we should use `getInstance` method for singleton-like classes instead of MetaClass magic. When a developer calls `getInstance` they understand that they are working with the global state. While in the case of metaclasses they are not.

emilmelnikov:
> I think we should use `getInstance` method for singleton-like classes instead of MetaClass magic. When a developer calls `getInstance` they understand that they are working with the global state. While in the case of metaclasses they are not.

I would say that singletons should not be used at all because Python has module-level variables.

m-novikov:
> I would say that singletons should not be used at all because Python has module-level variables.
Not sure how it would be solved when we have multiple PreferencesManagers. Btw, why do we need preferences managers with different paths?

emilmelnikov:
> Not sure how it would be solved when we have multiple PreferencesManagers. Btw, why do we need preferences managers with different paths?

Do we really need multiple preferences managers? I am currently trying to replace `PreferencesManager` machinery with a bunch of functions in a module, and the only reason to have multiple preferences is for testing, but this could be solved by temporarily switching the preferences location to some different temporary file.

Tomaz-Vieira:
I think I agree with everything said here... there doesn't seem to be a reason to be able to create multiple `PreferencesManagers`. The only thing is I don't quite like is the `get_location` and `set_location` functions... Maybe we shouldn't be setting this programatically at all; just use the default path or read an override out of an environment variable on startup time and then never again allow it to be changed. Otherwise we might stumble into some weird behavior if someone tries to change the paths while ilastik is already running

emilmelnikov:
> The only thing is I don't quite like is the `get_location` and `set_location` functions... Maybe we shouldn't be setting this programatically at all; just use the default path or read an override out of an environment variable on startup time and then never again allow it to be changed.

These have been added "mocking" the contents of the preferences file in tests. I would say that they are OK to have even without tests, because, for example, if we read the location from environment variable, then we need a setter function.

> Otherwise we might stumble into some weird behavior if someone tries to change the paths while ilastik is already running

All `get*` and `set*` functions are transactional, and also mutex-protected, so they should be safe to use within a single process. If someone tries to mess with the file outside our process, all bets are off, but we cannot do anything about this anyway.

Tomaz-Vieira:
> All `get*` and `set*` functions are transactional, and also mutex-protected, so they should be safe to use within a single process. If someone tries to mess with the file outside of our process, then all bets are off, but we cannot do anything about this anyway.

I'm not even worried about thread-safety. I'm worried that some part of ilastik might switch the path to the preferences file and we might end up with some parts of the preferences being saved in one file and some parts being saved in another. This is almost equivalent to having multiple instances of `PreferencesManager`, which is why I think we might want to NOT allow changing that path in runtime at all.

emilmelnikov:
> > All `get*` and `set*` functions are transactional, and also mutex-protected, so they should be safe to use within a single process. If someone tries to mess with the file outside of our process, then all bets are off, but we cannot do anything about this anyway.
> 
> I'm not even worried about thread-safety. I'm worried that some part of ilastik might switch the path to the preferences file and we might end up with some parts of the preferences being saved in one file and some parts being saved in another. This is almost equivalent to having multiple instances of `PreferencesManager`, which is why I think we might want to NOT allow changing that path in runtime at all.

In this case, how would you test functions in the `preferences` module without touching the default file?

Tomaz-Vieira:
> In this case, how would you test functions in the `preferences` module without overwriting the default file?

Hmm what do you think of environment variables? It would be good even for developer purposes if we could run ilastik like `PREFS_FILE=/tmp/whatever.prefs python ilastik.py`

Or maybe a thin child class, provided via a fixture, that just overrides the path for testing purposes?

You might have some better ideas... main point is i think it's good to not provide utilities that are not supposed to be used; I mean, that's the main reason why we're moving away from being able instantiate `PreferencesManager` =D

emilmelnikov:
> Hmm what do you think of environment variables? It would be good even for developer purposes if we could run ilastik like `PREFS_FILE=/tmp/whatever.prefs python ilastik.py`
> 
> Or maybe a thin child class, provided via a fixture, that just overrides the path for testing purposes?
> 
> You might have some better ideas... main point is i think it's good to not provide utilities that are not supposed to be used; I mean, that's the main reason why we're moving away from being able instantiate `PreferencesManager` =D

I think the main argument against `PreferencesManager` is it's singleton nature: people usually expect a new object to be returned when they call `__init__`.

Regarding the ability to change location: I think it won't be a problem. Changing preferences location makes sense only during application startup (e.g. when reading the location from an environment variable) or in tests. Sure, someone might change the location deep in the code, but (1) I would assume that we always do code reviews, and (2) it is implausible that someone calls `set_location` accidentally. At the same time, other solutions look too complicated for such a simple problem.

Tomaz-Vieira:
Close din favor of #260

PR #258, opened 2019-09-26
Move special case for mouse events to HUD

emilmelnikov:
Previously, EventSwtich installed itself as an event filter to both
ImageView2D (which is QGraphicsView) and it's viewport because a single
event filter should be able to catch all events (see https://stackoverflow.com/questions/2445997/qgraphicsview-and-eventfilter
for details). This made it hard to understand how events interact with
different event filters.

Another solution described in one of the anwsers in the link above is to
remove focus proxy from QGraphicsView. QAbstractScrollArea (superclass
of QGraphicsView) installs itself as a focus proxy to it's viewport
in order to be able to handle scrolling via keyboard. Since our
application doesn't need that, it is safe to remove the focus proxy and
simplify event handling.

However, HUD is currently displayed as a sibling of QGraphicsView
viewport, which causes problems when a user wants to do something with
a mouse in a HUD area that doesn't contain buttons, but has a usable
area underneath. We solve that problem by sending all mouse events from
HUD to the viewport of QGraphicsView.

Closes https://github.com/ilastik/ilastik/issues/2091.

### DIFF
--- a/volumina/eventswitch.py
+++ b/volumina/eventswitch.py
@@ -20,8 +20,10 @@
 # 		   http://ilastik.org/license/
 ###############################################################################
 
-from PyQt5.QtCore import QObject, QEvent
+from typing import Iterable
 
+from PyQt5.QtCore import QEvent, QObject
+from PyQt5.QtWidgets import QGraphicsView
 from volumina.utility.qabc import QABC, abstractmethod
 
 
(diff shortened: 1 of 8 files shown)


volumina/eventswitch.py
```diff
@@ -92,26 +70,13 @@ def __init__(self, imageviews, interpreter=None):
...
+        for imageview in imageviews:
+            # QAbstractScrollArea (QGraphicsView superclass) installs itself as a focus proxy into a viewport.
+            # Since we want to receive key events in the event filter, uninstall the focus proxy from the viewport.
+            # See https://stackoverflow.com/a/2501489.
+            viewport = imageview.viewport()
+            viewport.setFocusProxy(None)
+            viewport.installEventFilter(self)
+
+    def eventFilter(self, watched: QObject, event: QEvent) -> bool:
+        return self._interpreter.eventFilter(watched.parent(), event)
```
k-dominik: `self._interpreter` should either not have a default value of `None` or not be optional in this class. Or one should check in this method whether it is `None`, but that doesn't seem to make a lot of sense.

emilmelnikov:
> once you confirmed this is working on windows feel free to merge it :)

Works on Windows.

PR #260, opened 2019-09-30
Replace PreferencesManager with module functions

emilmelnikov:
PreferencesManager is a singleton that controls access to a preferences
file on disk. Since Python has module-level variables, singleton classes
can be replaced with private variables and public functions in a module.

See also: https://github.com/ilastik/volumina/pull/257.


volumina/utility/preferences.py
```diff
@@ -0,0 +1,155 @@
...
+    """Change the preferences file location.
+
+    All subsequent preferences operations will use the specified file.
+    This function simply changes the source for reads and writes, it
+    does not copy the contents of the previous file into a new one.
+
+    Returns:
+        Old preferences file location.
+    """
+    with _lock:
```
Tomaz-Vieira: I think this `_lock` is not doing much locking at all... It doesn't protect a client from multiple other clients changing the `_path`. Say we have two clients like so:

```
preferences.change_location("first_location") # client 1
preferences.set("some_group", "some_property", 123) # client 1
```

```
preferences.change_location("second_location") #client 2
preferences.set("some_group", "some_property", 456) #client 2
```

then this is a valid execution order:

```
preferences.change_location("first_location") # client 1
preferences.change_location("second_location") #client 2
preferences.set("some_group", "some_property", 123) # client 1
preferences.set("some_group", "some_property", 456) #client 2
```
and even with the locks, everything would be written into `"second_location"`. I see that this has somewhat of an "exchange" semantics, but for that to work the return (old) value would have to have a value that indicates whether you acquired a lock.

Anyway, it might be that we don't even care about having this thread-safe anyway, so we might just do away with the locks altogether.

emilmelnikov: You are right, but I concluded that it would be safer to serialize all accesses to preferences anyway because contention shouldn't be a problem: if you read preferences in a tight loop, then you are doing something wrong.

> Anyway, it might be that we don't even care about having this thread-safe anyway, so we might just do away with the locks altogether.

This is also correct, as @k-dominik pointed out, we can load data from a file and keep it in memory, while stating that `change_location` is not thread-safe.

emilmelnikov:
> I'd just wish we could get rid of pickling the file to make it human readable, but then we would not be able to jump between versions flawlessly anymore.

Initially I wanted to replace pickle with JSON, but that would require too much work for a little value.

> Little comment, why do you reload the settings for every "transaction"? I get that this makes it consistent over multiple instance, but the question for me here, is that what we want? Personally I would be frustrated if the last openend file path would leak between instances during the same session. And sure, the content would be out of whatever cached preferences we had would be out of sync, but would users care? If consistency becomes annoying, and I think it would in this instance, then we should probably not insist on it.

This is a very good point! We don't want that.

volumina/utility/preferences.py
```diff
@@ -0,0 +1,143 @@
...
+    If the group or the setting in a group does not exist, it will be
+    created.
+
+    See Also:
+        :func:`setmany`.
+    """
+    return _preferences.set(group, setting, value)
+
+
+def getmany(*args: Tuple[str, str, Any]) -> Tuple[Any, ...]:
```
m-novikov: I think as no other module except tests uses these methods (`getmany`, `setmany`), we should not make them public nor on module level, nor on class.

volumina/utility/preferences.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _load_preferences(path: pathlib.Path) -> MutableMapping[str, MutableMapping[str, Any]]:
+    try:
+        with open(path, "rb") as f:
+            return pickle.load(f)
+    except FileNotFoundError:
+        return {}
+    except Exception as e:
+        logger.exception(f"Failed to load preferences from {str(path)!r}")
+        if not isinstance(e, IOError):
```
m-novikov: Should we clarify why we are doing this in the comment and why we are checking for IOError?
Also if the parent folder of this file has only read permissions unlink will raise PermissionDenied.

volumina/utility/preferences.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class _Preferences:
+    def __init__(self, path: pathlib.Path):
+        self._path = path
+        self._data = _load_preferences(path)
+        self._lock = threading.Lock()
+
+    def get(self, group: str, setting: str, default: Any = None) -> Any:
+        return self.getmany((group, setting, default))[0]
```
m-novikov: Do we really need getmany?

volumina/utility/preferences.py
emilmelnikov: `setmany` is [used in liastik](https://github.com/ilastik/ilastik/pull/2135/files#diff-870f52209923664e0309e4912a5f5069R507).

volumina/utility/preferences.py
emilmelnikov: `IOError` means that we (probably) won't be able to unlink the file, hence the instance check.

volumina/utility/preferences.py
emilmelnikov: It is not used anywhere, but was added for the symmetry with `setmany`.

volumina/utility/preferences.py
m-novikov: Do we need a full function to avoid calling two consecutive sets?

emilmelnikov: Yes, because another thread might change one of the values between two calls.

PR #261, opened 2020-03-27
fixes crash related to Overview3d with Python 3.7.4 on OSX

k-dominik:
Since Python 3.7.4 an error bubbles up that has been ignored in previous Python versions.
The cause is `PyQt5` accessing all class attributes via `PyObject_GetAttr` during initialization of the Gui Widget. In our case, the class is not fully initalized by that time but class properties are still executed (via `PyObject_GetAttr`).

There are a couple of solutions to this problem ([related thread on riverbank mailinglist](https://www.riverbankcomputing.com/pipermail/pyqt/2019-August/042020.html)

 * this PR: properties removed in favor of get/set functions.
 * wait until PyQt5 5.13.1 is available on conda-forge -> the issue is fixed there
 * switch to PySide2 for the whole of ilastik. Those are the official Qt bindings.

### DIFF
--- a/volumina/navigationController.py
+++ b/volumina/navigationController.py
@@ -372,7 +372,7 @@ class NavigationController(QObject):
 
         def maybeUpdateSlice(oldSlicing):
             if oldSlicing == self._model.slicingPos:
-                self._view3d.slice = self._model.slicingPos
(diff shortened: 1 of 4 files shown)


k-dominik:
merging this in order to move on with qt 5.12

PR #263, opened 2020-04-17
source axistags from input, allow for singleton axes when changing order

k-dominik:
This PR allows the User to reoder axes _and_ keep singletons present in the original data

* sources the `inputAxes` from the input's original shape
* allow to have singleton axes (that were present in original shape) when
  reordering

### DIFF
--- a/volumina/widgets/dataExportOptionsDlg.py
+++ b/volumina/widgets/dataExportOptionsDlg.py
@@ -374,9 +374,7 @@ class DataExportOptionsDlg(QDialog):
 
         self.outputAxisOrderEdit.editingFinished.connect(_handleNewAxisOrder)
         self.outputAxisOrderEdit.textChanged.connect(partial(self._updateAxisOrderColor, True))
-        self.outputAxisOrderEdit.setValidator(
-            DataExportOptionsDlg._AxisOrderValidator(self, self._opDataExport.ConvertedImage)
-        )
+        self.outputAxisOrderEdit.setValidator(DataExportOptionsDlg._AxisOrderValidator(self, self._opDataExport.Input))
         self.outputAxisOrderEdit.installEventFilter(DataExportOptionsDlg._AxisOrderEventFilter(self))
         self.axisOrderCheckbox.toggled.connect(_handleAxisOrderChecked)
 
@@ -410,7 +408,8 @@ class DataExportOptionsDlg(QDialog):
 
         def validate(self, userAxes, pos):
             taggedShape = self._inputSlot.meta.getTaggedShape()
-            inputAxes = list(taggedShape.keys())
+            originalAxes = self._inputSlot.meta.getOriginalAxisKeys()
+            inputAxes = list(originalAxes)
             inputSet = set(inputAxes)
             userSet = set(str(userAxes))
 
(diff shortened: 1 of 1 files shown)


volumina/widgets/dataExportOptionsDlg.py
```diff
@@ -410,7 +408,8 @@ def __init__(self, parent, inputSlot):
 
         def validate(self, userAxes, pos):
             taggedShape = self._inputSlot.meta.getTaggedShape()
-            inputAxes = list(taggedShape.keys())
+            originalAxes = self._inputSlot.meta.getOriginalAxisKeys()
```
k-dominik, line 411: shouldn't we just allow users to create all the axes from `{t, c, z, y, x}` they want? @Tomaz-Vieira

k-dominik: okay I went ahead to allow all of those when specifying output. There is not reason users shouldn't be able to export 5D if they choose to.

volumina/widgets/dataExportOptionsDlg.py
```diff
@@ -415,7 +416,7 @@ def validate(self, userAxes, pos):
             userSet = set(str(userAxes))
 
             # Ensure all user axes appear in the input
-            if not (userSet <= inputSet):
+            if not inputSet.issubset(self._valid_axes):
```
Tomaz-Vieira, line 419: Shouldn't this be `if not userSet.issubset(self.__valid_axes)`?  Unless there's a weird bug elsewhere in ilastik, `inputSet` will always be a subset of `"tzyxc"`

Also, maybe fix the comment to something that reflects the change like "Ensure all axes are valid" =)

volumina/widgets/dataExportOptionsDlg.py
```diff
@@ -425,7 +426,7 @@ def validate(self, userAxes, pos):
             # If missing non-singleton axes, maybe intermediate entry
             # (It's okay to omit singleton axes)
             for key in inputSet - userSet:
-                if taggedShape[key] != 1:
+                if taggedShape.get(key, 1) != 1:
```
Tomaz-Vieira, line 429: I think this for-loop used to mean "For every input axis not mentioned by the user, make sure that it's a singleton". This change with the `.get[key, 1]` instead of `[key]` doesn't affect that, because `key` is always in `taggedShape` anyway.

If the user entered some axes that are not present in `inputSet` (and we have the necessary magic to interpret that as singletons in the export logic), then there's no further checking we need to do after validating that `userSet.issubset(self.__valid_axes)` and that there are no repeated axes

Tomaz-Vieira (changes requested):

volumina/widgets/dataExportOptionsDlg.py
k-dominik: ah yeah. I think naming here is suboptimal to invite input / user (user-input?!) confusion ;)

k-dominik:
good that you had a look. I think I see the problem more clearly now. Somewhere in ilastik, some operator is dropping singleton axes... let's see where that will end up.

for this PR, I'll make sure users can add all the singletons they want, nothing else

PR #267, opened 2020-09-04
Save preferences in JSON format instead of pickle

emilmelnikov:
Closes https://github.com/ilastik/ilastik/issues/2300.


m-novikov:
Great :) Thanks for looking into it.
I think it would be nice if no json preferences have been found try to migrate them from pickled version

volumina/utility/preferences.py
```diff
@@ -25,75 +25,73 @@
...
-        return {}
+    def _write(self, data):
+        with open(self.path, "w") as f:
+            json.dump(data, f, indent=2, sort_keys=True)
 
 
-_preferences = _Preferences(pathlib.Path.home() / ".ilastik_preferences")
+_preferences = Preferences(
+    os.path.join(
+        appdirs.user_config_dir(appname="ilastik", appauthor=False),
```
k-dominik, line 91: I see you're moving the config away from just home. It is good to "respect" the OS's wishes here. But this also comes with the downside that you cannot simply point users to their home folders, but need to remember (check) where the specific config location are on the specific OS. So I'd rather keep it in home.

k-dominik (commented):
this is great, but I second the notion for trying to import from the old config file if no preference file was found.

volumina/utility/preferences.py
emilmelnikov: The better approach may be adding a menu item "Reveal in (Finder|Explorer|etc.)" to ilastik GUI that opens a config folder.

emilmelnikov: We can also use this folder to keep other ilastik stuff.

emilmelnikov:
The migration should be triggered from somewhere. Should we do this in `ilastik.app.main`?

volumina/utility/preferences.py
k-dominik: > The better approach may be adding a menu item "Reveal in (Finder|Explorer|etc.)" to ilastik GUI that opens a config folder.

Just a story from the past: ilastik wouldn't show up for users, thought it clearly started. Turned out geometry was off, so the window was spawned somewhere in nirvana. Deleting the preferences solved the issue. A "reveal in whatever" function wouldn't have been helpful there.

k-dominik: but yeah, this is a fringe case. They certainly look better somewhere not in `$HOME`, since it is not something the user should play around with anyway.

emilmelnikov: > Just a story from the past: ...

That's a good point!

emilmelnikov: So, what should we do? I still think it's nice to keep configs and other data in the proper directories, and tell users in the documentation where to look if they stumble upon some obscure bug.

k-dominik (approved):
nice one!

volumina/utility/preferences.py
k-dominik: nah, that's fine :)

volumina/utility/preferences.py
```diff
@@ -25,75 +25,89 @@
...
+        old_path.unlink()
+        self.write(old_preferences)
+
+
+_preferences = Preferences(
+    os.path.join(
+        appdirs.user_config_dir(appname="ilastik", appauthor=False),
+        "preferences.json",
+    )
+)
```
Tomaz-Vieira, line 110: Seems like this would be a good place to have the several Paths where the config file could reside if we were to allow for that kind of thing

volumina/utility/preferences.py
```diff
@@ -25,75 +25,89 @@
...
+                return json.load(f)
+        except FileNotFoundError:
+            return {}
+
+    def write(self, data):
+        os.makedirs(self.path.parent, exist_ok=True)
+        with open(self.path, "w") as f:
+            json.dump(data, f, indent=2, sort_keys=True)
 
+    def migrate(self, old_path):
```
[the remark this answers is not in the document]
emilmelnikov: `Preferences` is just a wrapper for `path` under a mutex (it is easier to test than the previous approach with 2 module-level variables), so you need to use a single instance in order to keep the mutex around. But we don't claim the mutex when changing paths, which should be fixed!

volumina/utility/preferences.py
```diff
@@ -25,75 +25,89 @@
...
 
+        try:
+            with open(old_path, "rb") as f:
+                old_preferences = pickle.load(f)
+        except Exception:
+            return
 
-_preferences = _Preferences(pathlib.Path.home() / ".ilastik_preferences")
+        old_path.unlink()
+        self.write(old_preferences)
```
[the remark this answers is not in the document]
emilmelnikov: Good point.

`Preferences.write` already flushes because it uses a contextmanager on a file handle inside.

volumina/utility/preferences.py
emilmelnikov: I don't think we need multiple locations for `preferences.json` since is is created automatically by volumina, and users won't see it under normal circumstances.

k-dominik:
LGTM!

k-dominik:
Hi @emilmelnikov,

do you want to do some rebasing here, or can I merge it as is?

emilmelnikov:
> Hi @emilmelnikov,
> 
> do you want to do some rebasing here, or can I merge it as is?

Do we need to rebase? Squash-merge should be fine.

k-dominik:
> > Hi @emilmelnikov,
> > do you want to do some rebasing here, or can I merge it as is?
> 
> Do we need to rebase? Squash-merge should be fine.

Do you want to collapse all changes to a single commit for this one?

emilmelnikov:
> > > Hi @emilmelnikov,
> > > do you want to do some rebasing here, or can I merge it as is?
> > 
> > 
> > Do we need to rebase? Squash-merge should be fine.
> 
> Do you want to collapse all changes to a single commit for this one?

Yes, unless there is a reason to do something different.

k-dominik:
> Yes, unless there is a reason to do something different.

preserve your well thought out meaningful commit messages?

k-dominik:
windows build fails because of old image: see https://github.com/ilastik/volumina/pull/273

PR #268, opened 2020-09-09
Cancel obsolete requests

m-novikov:


volumina/tiling/__init__.py
```diff
@@ -315,6 +317,12 @@ def requestRefresh(self, rectF, stack_id=None, prefetch=False, layer_indexes=Non
         """
         stack_id = stack_id or self._current_stack_id
         tile_nos = self.tiling.intersected(rectF)
+        to_cancel = self._current_requests
+        self._current_requests = {}
```
Tomaz-Vieira, line 321: Volumina's code is kinda hard to navigate, so I'd love to see some type annotations like

`self._current_requests: Mapping[Tuple[LAYER_ID, ImageSource, REQUEST_ID], ArrayRequest] = {}`

volumina/tiling/__init__.py
```diff
@@ -552,6 +562,11 @@ def _fetch_layer_tile(self, timestamp, ims, transform, tile_nr, stack_id, ims_re
 
                 if stack_id == self._current_stack_id and cache is self._cache:
                     self.sceneRectChanged.emit(tile_rect)
+            else:
+                ims_req.cancel()
+
+            self._current_requests.pop((stack_id[1], ims, tile_nr), None)
```
Tomaz-Vieira, line 568: Are you using `pop` here  because the request might  have been already removed from `self._current_requests`?

This kinda makes me think if it would be possible for a request to overwrite an older one; If I, say, load an image in a layer, make a request, drop that layer, then load another image and make another request, is it possible for the second request to overwrite the previous one? Even worse, is it possible for the second one to be cancelled when we meant to cancel the first?

k-dominik:
closed in favor of #339

PR #269, opened 2020-09-19
Brushing example

m-novikov:


k-dominik (approved):
LGTM, always good to have some useful examples around :)

PR #270, opened 2020-09-19
Undo redo

m-novikov:
Demo with hotkeys Ctrl+Z for undo and Ctrl+Y for redo.
Example script for labeling in `examples/labeling.py`
undo-redo
Doesn't work for lazyflow slots.
Why do LF sinks have separate input and output?

### DIFF
--- a/volumina/pixelpipeline/datasources/arraysource.py
+++ b/volumina/pixelpipeline/datasources/arraysource.py
@@ -86,7 +86,7 @@ class ArraySinkSource(ArraySource):
 
 class RelabelingArraySource(ArraySource):
     """Applies a relabeling to each request before passing it on
-       Currently, it casts everything to uint8, so be careful."""
+    Currently, it casts everything to uint8, so be careful."""
 
     isDirty = pyqtSignal(object)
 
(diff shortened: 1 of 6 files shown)


m-novikov:
Ilastik version
ilastik-undo-redo

k-dominik:
this is awesome!!! This is fun :) Should go in ASAP

volumina/pixelpipeline/datasources/arraysource.py
```diff
@@ -77,14 +79,14 @@ def put(self, slicing, subarray, neutral=0):
         assert len(slicing) == len(
             self._array.shape
         ), "slicing into an array of shape=%r requested, but the slicing object is %r" % (slicing, self._array.shape)
-        self._array[slicing] = np.where(subarray != neutral, subarray, self._array[slicing])
+        self._array[slicing] = subarray
```
m-novikov, line 82: Now we always write data that we previously read, so there is no "neutral" value.
Before
```python
src = array([0, 1, 1])
labels = array([0, 2, 0])
put_arr = array([0, 2, 0])
# result write only put_arr[1]
```
now
```python
src = array([0, 1, 1])
labels = array([0, 2, 0])
put_arr = array([0, 2, 1])
# result write whole put_arr
```

k-dominik: right, that is not necessary anymore with your new implementation you do

```python
redo_data = numpy.where(self.__labels != 0, self.__labels, self.__old_data)
```
Before, you could draw, without requesting the data first and still only overpaint on stroke.

We should carefully chat, which code paths might also go to `put`...

k-dominik (approved):
Should not withhold this from our users! Awesome stuff.

Who thought that you can really get away adding such a feature without touching one line in the ilastik code base! Nice separation there.

PR #273, opened 2020-10-08
fix tests

k-dominik:
with recent dependency updates and merged branches tests started to fail

(my bad, didn't merge master before)

reasons:
* newer QT version does not play with the old windows image used in appveyor
  see also https://github.com/ilastik/ilastik/commit/1ee52d2400facc5a925edada3faddb52adee20af
* pytest doesn't like test files with same base names
* datasource factory for LazyflowSource now returns a `CacheSource`


appveyor.yml
```diff
@@ -21,7 +22,7 @@ install:
   # replace with whatever version of volumina triggered the appveyor
   - ps: cp -recurse C:\projects\volumina c:\ilastik\ilastik-meta\volumina
   - timeout 10
-  - python ilastik/scripts/devenv.py create -n %ENV_NAME% -c ilastik-forge conda-forge -p ilastik-dependencies-no-solvers
+  - python ilastik/scripts/devenv.py create -n %ENV_NAME% -c ilastik-forge conda-forge defaults -p ilastik-dependencies-no-solvers
```
emilmelnikov, line 25: That is off-topic, but we probably can remove this script since lazyflow is gone, and `conda-develop` is used. It seems that we can replace this with 2 lines of `conda create -n %ENV_NAME% ...` and `conda develop -n %ENV_NAME%`.

volumina/pixelpipeline/datasources/__init__.py
```diff
@@ -39,6 +39,9 @@
 try:
     from .lazyflowsource import LazyflowSource, LazyflowSinkSource
 
-    __all__ += ["LazyflowSource", "LazyflowSinkSource"]
+    __all__ += ["LazyflowSource", "LazyflowSinkSource", "CacheSource"]
+    from .cachesource import CacheSource
```
emilmelnikov, line 43: Should we import try to everything optional first, and then extend `__all__`?

k-dominik:
where is the :partyparrot: when you need it?

PR #275, opened 2020-10-12
temporary fix: fail a little more gracefully on pref import fail

k-dominik:
this is _not_ the proper way to do it, we should still import all
preference properly (key bindings) https://github.com/ilastik/ilastik/issues/2323, and ignore deprecated ones https://github.com/ilastik/ilastik/issues/2322 that
also result in failures.

We need to fix this quickly for the release
This should of course be followed up properly.

This PR preserves the old file if import failed with the hope we'll properly fix it at some point.


k-dominik:
I'll merge this. @m-novikov also brought up the alternative of reverting the preference related PRs, but this is not possible to do clean.

PR #276, opened 2020-10-14
Save shortcuts as list of dicts

emilmelnikov:
Old pickled preferences saved shortcuts as a dict with tuple keys, but
JSON can only have string keys, so transform the old dict into the new
format.

Closes https://github.com/ilastik/ilastik/issues/2323

### DIFF
--- a/volumina/utility/preferences.py
+++ b/volumina/utility/preferences.py
@@ -106,16 +106,15 @@ class Preferences:
         from volumina.utility import ShortcutManager as SM
 
         try:
-            with open(old_path, "rb") as f:
-                old_preferences = pickle.load(f)
-                # Old pickled preferences saved shortcuts as a dict
-                # with tuple keys, but JSON can only have string keys,
-                # so transform the old dict into the new format.
-                reversemap = old_preferences.get(SM.PreferencesGroup, {}).get(SM.PreferencesSetting, {})
-                if reversemap:
-                    old_preferences.setdefault(SM.PreferencesGroup, {})[SM.PreferencesSetting] = [
-                        {"group": group, "name": name, "keyseq": keyseq} for (group, name), keyseq in reversemap.items()
-                    ]
+            old_preferences = pickle.loads(old_path.read_bytes())
+            # Old pickled preferences saved shortcuts as a dict
+            # with tuple keys, but JSON can only have string keys,
+            # so transform the old dict into the new format.
+            reversemap = old_preferences.get(SM.PreferencesGroup, {}).get(SM.PreferencesSetting, {})
+            if reversemap:
+                old_preferences.setdefault(SM.PreferencesGroup, {})[SM.PreferencesSetting] = [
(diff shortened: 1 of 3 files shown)


volumina/utility/preferences.py
```diff
@@ -77,37 +77,50 @@ def setmany(self, *args: Tuple[str, str, Any]) -> None:
...
+        try:
+            with self._lock:
+                os.makedirs(self.path.parent, exist_ok=True)
+                # Touch the file only after successful serialization
+                # in order to prevent existing file corruption.
+                self.path.write_text(json.dumps(data, indent=2, sort_keys=True))
+            return True
+        except Exception:
+            logger.exception("Failed to write preferences to %s", self.path)
+            self.path.unlink()
```
k-dominik, line 98: only small issue, when deleting the file, we will not be able to investigate the cause easily. Maybe rename to `prefs.broken` or something?

volumina/utility/preferences.py
```diff
@@ -77,37 +77,50 @@ def setmany(self, *args: Tuple[str, str, Any]) -> None:
...
+                ]
         except Exception:
-            logger.warning("Encountered an exception on preferences import - skipping import.")
+            logger.exception("Failed to read old preferences")
             return
 
-        self.write(old_preferences)
-        old_path.unlink()
+        if self.write(old_preferences):
+            old_path.unlink()
```
k-dominik, line 123: Also here, renaming to `ilastik_preferences.old` or something would be nice, in case users want to go back to the previous versions (I am constantly jumping through our versions, and with  this I loose all my settings for the old version).

k-dominik (approved):
Very nice! Works as expected :+1:

volumina/utility/preferences.py
emilmelnikov: How about just leaving the file as-is? This can fail only if there is a serialization error (then we should indeed just leave the filesystem alone), or an I/O error (this should be rare, and there is no point in doing more I/O anyway).

volumina/utility/preferences.py
emilmelnikov: We can just leave the old file as-is here too. New versions read from a new locations anyway.

This way, new ilastik versions will look at the old location, transform/copy those preferences to the new location at startup, and use only the new location, while old versions will just use the old location as if nothing happened.

k-dominik: I thought about this more and looking at the code, this does not seem to be the case. `migrate` is always executed when starting ilastik. This should _only_ be done if there is no _new_ preferences file. If we do it like that, the two can coexist in isolation, which I think is a good enough solution.

emilmelnikov: I've changed the migration _exit_ condition to `self.path.exists() and not old_path.exists()`, that is, we actually migrate only if we didn't do it before (there is no new file) and we have something to migrate from (the old file exists).

k-dominik: that's perfect! if you remove the test assertion that old file is gone, this can be merged!

k-dominik (approved):

PR #278, opened 2021-02-10
Make volumina a standalone conda package

k-dominik:
This will ease ilastik deployment and grant us a little bit of flexibility in the future.

Todo:

- [ ] add proper CI
- [x] add minimal cmd interface + entrypoint


.bumpversion.cfg
```diff
@@ -0,0 +1,15 @@
+[bumpversion]
```
emilmelnikov, line 1: Not sure about this due to our use of conda, but this can be replaced with [setuptools_scm](https://pypi.org/project/setuptools-scm/).

setup.py
```diff
@@ -23,26 +23,42 @@
...
 ]
 
 package_data = {
     "volumina.widgets": ["*.ui", "ui/*.ui"],
     "volumina": ["*.ui"],
-    "volumina._testing": ["*.tif", "*.png", "lena.npy", "rgba129x104.npy"],
+    "volumina._testing": ["*.tif", "*.png", "2d_cells_apoptotic_1channel.npy", "rgba129x104.npy"],
 }
 
 setup(
```
emilmelnikov, line 43: Most of the package metadata from `setup.py` can be moved to [declarative `setup.cfg`](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html).

.bumpversion.cfg
k-dominik: Last time i tried setuptools_scm (ages ago), I don't think I got a working conda recipe. I tend to like bumpversion. But yeah, one more step to version, than setuptools_scm, and potential for errors if not using bumpversion to commit.

k-dominik: okay, I got that working. It does not feel intrgrated with conda at all (had to use it in a hacky way). But at least there's only a single source of truth now :)

setup.py
k-dominik: it's so short, easily fits my screen. I'd rather consolidate setup.cfg with pyproject.toml, and leave setup.py as it is.

k-dominik:
thx Emil for the review! I'll close this PR in favor of #279

PR #279, opened 2021-02-11
Add Conda Recipe for volumina

k-dominik:
Superseeds #278 in order to trigger CI from this PR I reopened it from the main repo.

Todo:
- [x] check that deployment actually works

Summary:

* added a conda recipe for volumina, to make it a standalone package
* versioning via `setuptools_scm` (albeit a bit hacky)
* added a tiny entrypoint so that it does something useful on it's own
* updated Readme
* got rid of old ci, old dev stuff that was not used anyways.
* switched ci to github actions.

I triggered the build that will only run for tags on master from now on; the package can be tested via

```bash
conda install -c ilastik-forge -c conda-forge volumina
```


setup.py
```diff
@@ -20,29 +20,53 @@
...
-    "volumina.widgets": ["*.ui", "ui/*.ui"],
     "volumina": ["*.ui"],
-    "volumina._testing": ["*.tif", "*.png", "lena.npy", "rgba129x104.npy"],
+    "volumina._testing": ["*.npy", "*.txt"],
+    "volumina.icons": ["*"],
+    "volumina.view3d": ["ui/*.ui"],
+    "volumina.widgets": ["*.ui", "ui/*.ui"],
 }
 
+_version = setuptools_scm.get_version(write_to="volumina/_version.py")
```
emilmelnikov, line 47: When configuring a "normal" Python package with `setuptools_scm`, it's possible to omit `version` from `setup.py`/`setup.cfg` entirely, and do `__version__ = importlib.metadata.version("volumina")` in `volumina/__init__.py` (for Python < 3.8 there is a [backport](https://pypi.org/project/importlib-metadata/)).

Would this work with the conda setup?

volumina/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+        signal.signal(signal.SIGINT, lambda *_: app.quit())
+        app.exec_()
+    finally:
+        app.quit()
+
+
+def axiorder_type(value):
+    reasons = []
+    if len(set(value)) != len(value):
+        reasons.append("Each axis may only appear ones")
```
emilmelnikov, line 33:
```suggestion
        reasons.append("Each axis may only appear once")
```

volumina/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def parse_args():
+    p = argparse.ArgumentParser(
+        description="",
+        usage="",
+        epilog="",
+    )
+    p.add_argument("image", help="Path to .npy image")
+    p.add_argument("axistags", help="Strings describing axes in image. Valid values: 'tzyxc'", type=axiorder_type)
```
emilmelnikov, line 52: Maybe this argument should be an option? Some other type of `image` could have the axis info in the future.

volumina/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+    p.add_argument("--version", action="version", version=__version__)
+
+    args = p.parse_args()
+    return args
+
+
+def reorder_to_volumina(data, axistags):
+    tagged_data = xarray.DataArray(data, dims=tuple(axistags))
```
emilmelnikov, line 60: Are we depending on `xarray` just for axis reordering?

If this is the case, then it's trivial to do this ourselves, and we can drop the dependency.

```python
def normalize_axes(a, axes, *, order=('t', 'x', 'y', 'z', 'c')):
    transpose_axes = [axes.index(d) for d in order if d in axes]
    expand_axes = [i for i, d in enumerate(order) if d not in axes]
    return numpy.expand_dims(numpy.transpose(a, transpose_axes), expand_axes)
```

setup.py
k-dominik: I didn't get this working. Really it was the only way to duplicate it.

volumina/__main__.py
k-dominik: yeah, let's decide about this in the future then (of course you are correct, I already have a branch to read ilasitk h5 output that already has axistags :))

volumina/__main__.py
k-dominik: I'd like to keep it (at some point adding an xarray source). The only dependency that might be cumbersome might be pandas that is brought in via xarray.
What reservation towards xarray do you have (we depend on it in ilastik already)?

emilmelnikov: I like xarray, but I'm against adding dependencies for doing trivial things. However, in this case it's probably OK if we are planning to use it more.

volumina/__main__.py
emilmelnikov: I was suggesting the following:
```suggestion
    p.add_argument("--axistags", help="Strings describing axes in image. Valid values: 'tzyxc'", type=axiorder_type, required=True)
```

Later, we'll drop `required=True` and decide on an image type basis.

volumina/__main__.py
k-dominik: I want to strongly endorse using it, so yep, I'd like to keep it in here.

setup.py
```diff
@@ -20,29 +20,53 @@
...
     package_data=package_data,
-    setup_requires=["nose>=1.0"],
+    python_requires=">=3.7",
+    install_requires=[
+        "appdirs",
+        "cachetools",
+        "future",
+        "h5py",
+        "pyopengl",
+        "pyqt>=5.6",
```
m-novikov, line 63: should be pyqt5

setup.py
```diff
@@ -20,29 +20,53 @@
...
+        "pyqtgraph",
+        "qimage2ndarray",
+        "typing_extensions",
+        "vigra",
```
m-novikov, line 67: no vigra in pip, so no reason to specify it here I suppose

volumina/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@contextlib.contextmanager
+def volumina_viewer():
+    app = QApplication(sys.argv)
+    v = Viewer()
+    v.editor.setInteractionMode("navigation")
+    yield v
+    try:
+        signal.signal(signal.SIGINT, lambda *_: app.quit())
```
m-novikov, line 24: There is actually a bug when ctrl+c is pressed when no lazyflow is installed. Due to the incorrect override of protected `_work_queue` in `prioritizedThreadPool.py`. I don't think it's super relevant for this PR, but it would be nice to track this with an issue.

setup.py
k-dominik: hmmm in this case I think I'll move reqs out of setup.py to meta.yaml. I kept it here to have a _single_ location for deps.

volumina/__main__.py
k-dominik: geee, would you care to open this issue?

setup.py
k-dominik: in conda it's pyqt - will move deps to meta.yaml

volumina/__main__.py
m-novikov: https://github.com/ilastik/ilastik/issues/2399

PR #280, opened 2021-02-15
Set tilewidth automatically for 2D and 3D data

k-dominik:
I cannot believe that we did not have an issue about this open @akreshuk, but I couldn't find one.
With this PR we add a new preference key `"tileWidth3D"` which will be used to set tile width for 3D data. The preference key `"tileWidth"` remains and will be used for 2D data. Defaults are 256 for 3D, 512 for 2D.
I also moved responsibility for the preference handling from `ImageScene2D` to `VolumeEditorWidget`.

Added the setFocus so this PR actually closes an issue ;)
fixes ilastik/ilastik#2260

### DIFF
--- a/volumina/volumeEditorWidget.py
+++ b/volumina/volumeEditorWidget.py
@@ -59,6 +59,7 @@ from .sliceSelectorHud import ImageView2DHud, QuadStatusBar
 from .pixelpipeline.datasources import ArraySource
 from .volumeEditor import VolumeEditor
 from volumina.utility import ShortcutManager
+from volumina.utility import preferences
 
 
 class __TimerEventEater(QObject):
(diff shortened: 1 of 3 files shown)


volumina/volumeEditorWidget.py
```diff
@@ -324,6 +327,33 @@ def _rubberBandZoom(self):
                     self.editor.imageViews[self.editor._lastImageViewFocus]._cursorBackup
                 )
 
+    def _updateTileWidth(self):
+        tile_width = self._getTileWidth()
+        self.editor.setTileWidth(tile_width)
+
+    def _getTileWidthConfigKeyDefault(self):
+        singletons_spacial = [True for dim in self.editor.posModel.shape if dim == 1]
```
emilmelnikov, line 335:
```suggestion
        singletons_spacial = sum(1 for dim in self.editor.posModel.shape if dim == 1)
```

akreshuk:
Hallelujah!

PR #286, opened 2021-09-25
Fix clickable area on higher zoom levels

m-novikov:
Fixes https://github.com/ilastik/ilastik/issues/2476
The original issue with multicat is that clickable are doesn't match visual representation on high zoom levels which results in collisions and prevents clicking on certain segmentation edges.
This patch takes into account the scaling factor of the QGraphicsView and adjusts the clickable area accordingly. See screenshots with left is with this patch, right is without.

clickable_area


m-novikov:
Added some interactivity for users to see what they are actually clicking. A bit convoluted as it's possible multiple edges to have same id

file

k-dominik:
Hey @m-novikov,

this is super awesome! I checked it out - fun to use and works flawlessly in 3D, too.

Some comments, observations:

* ideally color would not change - in multicut we already have many colors for the edges...
* this is automatically activated for all `SegmentationEdgesLayer`. I think ideally this would only be on the Labels layer (`LabelableSegmentationEdgesLayer`). But there the problem would be that the unlabeled edges don't even have a color...
* Ideally the behavior could be at least turned on and off via config (but this I can also add later).

m-novikov:
https://user-images.githubusercontent.com/5163640/134804543-cc47f230-8dd8-4cdf-9c33-8a19aa307197.mp4

m-novikov:
> this is automatically activated for all SegmentationEdgesLayer. I think ideally this would only be on the Labels layer (LabelableSegmentationEdgesLayer). But there the problem would be that the unlabeled edges don't even have a color...

Color for edges https://github.com/ilastik/ilastik/pull/2479 and disabled hover events for the non labeling edges

> Ideally the behavior could be at least turned on and off via config (but this I can also add later).

Unsure where it should be configured `.voluminarc`?

k-dominik:
maybe it makes sense to separate PRs - one that "only" fixes the scaling of the clickable area, and a second one for highlighting?

m-novikov:
https://user-images.githubusercontent.com/5163640/134809508-0148ac66-23a0-442a-b4b3-b8445a5930be.mp4

k-dominik:
wow this looks super cool. :heart_eyes:

PR #287, opened 2021-09-26
Fix clickable area on higher zoom levels

m-novikov:
Fixes ilastik/ilastik#2476
Split from https://github.com/ilastik/volumina/pull/286


k-dominik (approved):
awesome! Thank you very much for this improvement :)

PR #293, opened 2022-05-29
Remove signed int export options

btbest:
Yay for necromancing old issues! I believe this is what was meant by https://github.com/ilastik/ilastik/issues/1934 - which would imply that the issue is on the wrong repo 😁 

The solution for adding tooltips feels ugly, any better suggestions? I also feel like there should be better phrasings for the tooltips, depending on the target audience. This was a first best guess.

### DIFF
--- a/volumina/widgets/dataExportOptionsDlg.py
+++ b/volumina/widgets/dataExportOptionsDlg.py
@@ -223,16 +223,12 @@ class DataExportOptionsDlg(QDialog):
         self.convertDtypeCheckbox.toggled.connect(_handleConvertDtypeChecked)
         dtypes = collections.OrderedDict(
             [
-                ("unsigned 8-bit", "uint8"),
-                ("unsigned 16-bit", "uint16"),
-                ("unsigned 32-bit", "uint32"),
-                ("unsigned 64-bit", "uint64"),
-                ("signed 8-bit", "int8"),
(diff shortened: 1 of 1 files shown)


volumina/widgets/dataExportOptionsDlg.py
```diff
@@ -221,29 +221,19 @@ def _handleConvertDtypeChecked():
...
+            ("integer 16-bit", "uint16", "Stores data as integers. More bits per value increase precision but require more storage and processing time."),
+            ("integer 32-bit", "uint32", "Stores data as integers. More bits per value increase precision but require more storage and processing time."),
+            ("integer 64-bit", "uint64", "Stores data as integers. More bits per value increase precision but require more storage and processing time."),
+            ("floating-point 32-bit", "float32", "Stores data as numbers with a decimal point. More bits per value increase precision but require more storage and processing time."),
+            ("floating-point 64-bit", "float64", "Stores data as numbers with a decimal point. More bits per value increase precision but require more storage and processing time."),
         ]
 
-        for name, dtype in list(dtypes.items()):
+        for i in range(0, len(dtypes)):
+            name, dtype, tooltip = dtypes[i]
```
emilmelnikov, line 234:
```suggestion
        for i, (name, dtype, tooltip) in enumerate(dtypes):
```

PR #294, opened 2022-06-03
Fix Scrolling: Alt + mouse wheel

jonashund:
Fix issue ilastik/ilastik#2009

Qt.AltModifier inverts vertical and horizontal scrolling directions as an [intended behaviour](https://bugreports.qt.io/browse/QTBUG-30948?focusedCommentId=202503&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-202503)

The change in indentation in line 36 was caused by `hook id: black` while committing.
Thus, I refrained from reverting it this time, but appreciate tips how to avoid such changes in the first place in the future.

## Checklist

- [x] Format code and imports.
- [x] Add comment.

### DIFF
--- a/volumina/navigationController.py
+++ b/volumina/navigationController.py
@@ -21,6 +21,7 @@
 ###############################################################################
 
 import copy
+import sys
 from functools import partial
 
 import volumina
@@ -33,7 +34,7 @@ from volumina.sliceIntersectionMarker import SliceIntersectionMarker
 
 def posView2D(pos3d, axis):
     """convert from a 3D position to a 2D position on the slicing plane
-       perpendicular to axis"""
+    perpendicular to axis"""
     pos2d = list(copy.deepcopy(pos3d))
     del pos2d[axis]
     return pos2d
@@ -148,7 +149,10 @@ class NavigationInterpreter(QObject, InterpreterABC):
         grviewCenter = imageview.mapToScene(imageview.viewport().rect().center())
 
         sign = 1
-        if event.angleDelta().y() < 0:
+        # Qt.AltModifier maps vertical (y) to horizontal (x) scrolling direction
+        if sys.platform != "darwin" and (k_alt or k_shift_alt) and event.angleDelta().x() < 0:
+            sign = -1
+        elif event.angleDelta().y() < 0:
             sign = -1
 
         if k_shift_alt:


volumina/navigationController.py
```diff
@@ -33,7 +33,7 @@
 
 def posView2D(pos3d, axis):
     """convert from a 3D position to a 2D position on the slicing plane
-       perpendicular to axis"""
+    perpendicular to axis"""
```
k-dominik, line 36: if our hooks do it then it's fine ;). I guess on our side a better way to handle this would be pinning black in the pre-commit hook to some specific version (black is still evolving...).

volumina/navigationController.py
```diff
@@ -148,7 +148,10 @@ def onWheel_default(self, imageview, event):
         grviewCenter = imageview.mapToScene(imageview.viewport().rect().center())
 
         sign = 1
-        if event.angleDelta().y() < 0:
+        if (k_ctrl or k_shift) and event.angleDelta().y() < 0:
+            sign = -1
+        elif (k_alt or k_shift_alt) and event.angleDelta().x() < 0:
+            # Qt.AltModifier maps vertical (y) to horizontal (x) scrolling direction
```
k-dominik, line 154:
```suggestion
        # Qt.AltModifier maps vertical (y) to horizontal (x) scrolling direction
        if sys.platform != "darwin" and (k_alt or k_shift_alt) and event.angleDelta().x() < 0:
            sign = -1
        elif event.angleDelta().y() < 0:
            sign = -1
```

the link to the bugtracker was helpful, as for me on a mac, the patch did the opposite: from working on the main branch to broken with the PR :). So I'd suggest to do something like the above. One could also use `platform.system()` to be more modern I guess, but `sys.platform` should suffice. (Note: code wasn't verified on Windows or Linux).

k-dominik (commented):
Hello @jonashund,

thank you for the fix! This was a really annoying bug! I'm currently using a mac, so I found  in my particular situation your code changes don't work, but I trust that your code should fix Windows and Linux. Thank you digging out that qt bugtracker, that was helpful as a reassurance. On what platform did you test?

The fix has to work on all platforms before we can accept it.

Cheers
Dominik

k-dominik:
as a bonus (for us) you could pin the qt version in the conda recipe to a max of `5.12.*`, so that the tests will pass :) (qt got a recently updated on conda-forge - haven't looked into why it's breaking volumina yet).

https://github.com/ilastik/volumina/blob/06b4b19eccb95ff54af13bc346587bbe2f2d2cd2/conda-recipe/meta.yaml#L23

conda-recipe/meta.yaml
```diff
@@ -20,7 +20,7 @@ requirements:
     - future
     - h5py
     - pyopengl
-    - pyqt >=5.6
+    - pyqt >=5.6,<=5.12.*
```
k-dominik, line 23:
```suggestion
    - pyqt >=5.6,<=5.12
```
this might be a conda/mamba bug (there will be a warning that the `.*` is superfluous and deprecated but I would still imagine it would work - it doesn't, that's why the tests didn't run through on that one

volumina/navigationController.py
```diff
@@ -148,7 +149,11 @@ def onWheel_default(self, imageview, event):
         grviewCenter = imageview.mapToScene(imageview.viewport().rect().center())
 
         sign = 1
-        if event.angleDelta().y() < 0:
+        # Qt.AltModifier maps vertical (y) to horizontal (x) scrolling direction
+        if sys.platform != "darwin" and (k_alt or k_shift_alt) and event.angleDelta().x() < 0:
+            sign = -1
+        elif event.angleDelta().y() < 0:
+            sign = -1
```
k-dominik, line 156:
```suggestion
```

`sign = -1` appears twice...

k-dominik (changes requested):
Hello @jonashund,

thank you for adding the pin! Unfortunately it looks like conda/mamba has a bug and won't solve the enviornment with `<=5.12.*`, I left a comment at the appropriate line. I really think after changing this the tests should pass and the PR is good to be merged. I have also tested it on my mac - it works :)

k-dominik (approved):

k-dominik:
thank you @jonashund for your contribution 🙌  :)

PR #296, opened 2022-09-21
Add button resetpath

sebherbert:
Hello all,

This small PR aims adds a reset button to the export gui so a user can revert to default "magic" path more easily. The corresponding ilastik PR can be found https://github.com/ilastik/ilastik/pull/2620

### DIFF
--- a/volumina/widgets/dataExportOptionsDlg.py
+++ b/volumina/widgets/dataExportOptionsDlg.py
@@ -137,7 +137,7 @@ class DataExportOptionsDlg(QDialog):
         # See self.eventFilter()
         self.installEventFilter(self)
 
-        def pressedResetButton():
+        def resetButtonClicked():
             opDataExport.OutputFilenameFormat.setValue(defaultExportPath)
             self.exportFileOptionsWidget.stackedWidget.currentWidget().updateFromSlots()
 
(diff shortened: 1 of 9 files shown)


volumina/widgets/dataExportOptionsDlg.py
```diff
@@ -135,6 +137,12 @@ def __init__(self, parent, opDataExport):
         # See self.eventFilter()
         self.installEventFilter(self)
 
+        def pressedResetButton():
```
k-dominik, line 140:
```suggestion
        def resetButtonClicked():
```

if we _have_ to do some bikeshedding I'd prefer to name it the other way round and also use the `clicked` signal which seems to be more standard, rather than the `pressed` signal.

k-dominik (commented):
Hello @sebherbert,

thank you so much for the PR :) Really looking forward to having this functionality in ilastik. I had a tiny nitbit related to naming and one observation:

At least on OSX the vertical distance from the filename widget to the button seems a bit excessive:

image

having said that, there was already weird spacing before:

image

so maybe a cosmetic change for the future (as unrelated to your PR)

volumina/widgets/dataExportOptionsDlg.py
sebherbert: Absolutely fine by me :)

sebherbert:
I'm stoked it worked out, looking forward to the next time I need it! :)

k-dominik:
Added a few commits to make the tests pass. Bit of a bummer that they will only pass in python 3.7, currently. I updated pins for the recipe, too and will trigger a new release of volumina once this one is merged...

k-dominik (approved):

k-dominik:
thx a lot for your contribution Seb!

PR #298, opened 2022-12-22
Add toggleTopToBottom utility function

k-dominik:
Added function to move a specific layer either to top of the layerstack (if it is at the bottom) or to the bottom (if anywhere else) So far only used for callbacks in layer shortcuts in ilastik (shortcut "i").

Co-Authored-By: Emil Melnikov <emilmelnikov@users.noreply.github.com>

### DIFF
--- a/volumina/layerstack.py
+++ b/volumina/layerstack.py
@@ -312,3 +312,16 @@ class LayerStackModel(QAbstractListModel):
             self[idx.row()].setActive(False)
         for idx in selected.indexes():
             self[idx.row()].setActive(True)
+
+    def toggleTopToBottom(self, layer: Layer) -> None:
+        """Select layer and move it to the top or bottom of the stack.
+
+        Layer goes to the top unless it's already there.
+        Top layer goes to the bottom.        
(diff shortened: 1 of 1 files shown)


volumina/layerstack.py
```diff
@@ -312,3 +312,11 @@ def _onSelectionChanged(self, selected, deselected):
             self[idx.row()].setActive(False)
         for idx in selected.indexes():
             self[idx.row()].setActive(True)
+
+    def toggleTopToBottom(self, target: Layer) -> None:
+        index = self.layerIndex(target)
```
emilmelnikov, line 317: Since it's somewhat hard to convey everything in the method's name, a brief docstring would be nice to have.

```suggestion
    def toggleTopToBottom(self, layer: Layer) -> None:
        """Select layer and move it to the top or bottom of the stack.

        Layer goes to the top unless it's already there.
        Top layer goes to the bottom.        
        """
        index = self.layerIndex(layer)
```

emilmelnikov (approved):

PR #299, opened 2023-01-23
Pin black version

emilmelnikov:
`pre-commit` complains when using a repo with branch name as a ref.

Also, add `*.egg-info` directory from `pip install -e` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.kate-swp
 .ropeproject
 *.DS_Store
+*.egg-info/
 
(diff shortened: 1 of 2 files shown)


k-dominik (approved):
great!

It would probably make sense to have the black versions in the pre-commit hook in sync accross ilastik and volumina (https://github.com/ilastik/ilastik/blob/d927177219f58736535dc01b0cf2f43e58252f20/.pre-commit-config.yaml#L3)

PR #304, opened 2023-08-31
enable QGL viewport

k-dominik:
Just tried this following our discussion whether `volumina` was using opengl during the meeting. Turns out this looks very promising - speedup in rendering is _very_ noticable. 

much faster rendering, especially in multicut with edge overlays.

enjoying some free lunch for now.

cc: @Tomaz-Vieira @emilmelnikov @btbest 

need to investigate some `QOpenGLFramebufferObject: Framebuffer incomplete attachment` "warnings" here and there...

### DIFF
--- a/volumina/config.py
+++ b/volumina/config.py
@@ -26,6 +26,7 @@ default_config = """
 [volumina]
 pixelpipeline_verbose: false
 show_3d_widget: true
+enable_fallback_viewports: false
 """
 
 _cfg = configparser.ConfigParser()
@@ -49,6 +50,10 @@ class _Config:
     @property
     def show_3d_widget(self):
         return self._get_boolean("volumina", "show_3d_widget")
+
+    @property
+    def enable_fallback_viewports(self):
+        return self._get_boolean("volumina", "enable_fallback_viewports")
 
     @property
     def cache_size(self):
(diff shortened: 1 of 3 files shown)


volumina/imageView2D.py
```diff
@@ -105,10 +104,8 @@ def __init__(self, parent, cropModel, imagescene2d):
 
         QGraphicsView.__init__(self, parent)
 
-        # We can't use OpenGL because the HUD doesn't render properly on top.
-        # Maybe this will be fixed in Qt5?
-        if False:
-            self.setViewport(QGLWidget())
+        # TODO: make this configurable
+        self.setViewport(QOpenGLWidget())
```
emilmelnikov, line 108: Use env var for that?..

```python
if os.environ.get("VOLUMINA_OPENGL", "0") == "1":
    self.setViewport(QOpenGLVidget())
```

volumina/config.py
```diff
@@ -26,6 +26,7 @@
 [volumina]
 pixelpipeline_verbose: false
 show_3d_widget: true
+disable_gl_viewports: false
```
emilmelnikov, line 29: Not sure about variable name here: double negation might be confusing (disable + false). On the other hand, this variable won't be used often, so maybe it's OK.

k-dominik: hehe, I first had it as `enable_gl_view`... would be more natural in code I agree, but setting env vars to `0` feels weird. Usually you check whether it is set or not...

PR #305, opened 2023-09-20
add support to load ilastik h5 flavors

k-dominik:
mostly because I'm annoyed i can not `volumina my-result-file.h5/exported_data` for quick viewing.

### DIFF
--- a/volumina/__main__.py
+++ b/volumina/__main__.py
@@ -1,8 +1,12 @@
 import argparse
 import contextlib
+import json
 import signal
 import sys
+from pathlib import Path
+from typing import Tuple, Union
 
+import h5py
 import numpy
 import xarray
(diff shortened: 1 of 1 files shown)


volumina/__main__.py
```diff
@@ -67,16 +79,93 @@ def reorder_to_volumina(data, axistags):
...
+
 def main():
     args = parse_args()
-    data = numpy.load(args.image)
-    reordered_data = reorder_to_volumina(data, args.axistags)
+    data, axistags = load(args.image)
+
+    if not (args.axistags or axistags):
+        print(f"Axistags required for file {args.image} with shape {data.shape}")
+        return 1
```
emilmelnikov, line 158: This will print a message to stderr and exit with code 1.

```python
sys.exit(f"Axistags required ...")
```

PR #307, opened 2023-10-26
stop hiding exceptions

k-dominik:
excepthook was used before in #251  2b09109b478a7f30787d9cecbad01ce82812a671 that claims segfaults being cause by it.

Could not reproduce segfaults - and rather have an error message popping up.
Otherwise in case of exception somewhere along the chain, ilastik looks like it's doing nothing and one wouldn't know.

### DIFF
--- a/volumina/tiling/__init__.py
+++ b/volumina/tiling/__init__.py
@@ -553,7 +553,7 @@ class TileProvider(QObject):
                 if stack_id == self._current_stack_id and cache is self._cache:
                     self.sceneRectChanged.emit(tile_rect)
         except BaseException:
-            logger.debug("Failed to fetch layer tile", exc_info=True)
+            raise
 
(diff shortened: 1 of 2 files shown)


volumina/tiling/__init__.py
```diff
@@ -553,7 +553,7 @@ def _fetch_layer_tile(self, timestamp, ims, transform, tile_nr, stack_id, ims_re
                 if stack_id == self._current_stack_id and cache is self._cache:
                     self.sceneRectChanged.emit(tile_rect)
         except BaseException:
-            logger.debug("Failed to fetch layer tile", exc_info=True)
+            sys.excepthook(*sys.exc_info())
```
k-dominik, line 556:
```suggestion
            raise
```

btbest (approved):

PR #310, opened 2024-01-11
Disallow bad pyqt5-sip versions

btbest:
This changes nothing that could make the pipeline fail.


btbest:
I can manually replay the steps of the test-w-ilastik-env action on my system and it works fine... so something off about github here I think

k-dominik (approved):
great that you tested those! Dedendency issues are not the most fun to solve...

PR #311, opened 2024-01-16
Refactor tiling

k-dominik:
In order to make it easier to extend tiling and also track changes this PR separates `Tiling` and `TileProvider` classes.

The only _addition_ here is the test for `Data2Scene` transformation in tiling.


btbest (approved):
Doesn't change any actual logic as far as I can tell so looks good. More tests are good tests as well.

PR #313, opened 2024-02-15
Prefer compressed formats for export

k-dominik:
ilastik outputs compress fairly well. We should encourage those per default.


btbest (approved):
ilastik's data export widget comes from volumina? 🙃

PR #316, opened 2024-09-18
Add export option for single-scale OME-Zarr

btbest:
Instead of adding a triplicate of the H5/N5 file options widget, I figured I'd combine them instead. Still pondering whether the one point of custom logic for zarr (allowing dataset to be empty) warrants a duplicate, but don't really think so. Not sure why it's mandatory anyway; there is ilastik code that enforces an internal path to be present for H5 but not N5.

I've put the menu option under the below hierarchical formats for now, but that's easy to relocate if we want to have it further up or down the list.


volumina/widgets/hierarchicalFileExportOptionsWidget.py
```diff
@@ -70,7 +74,7 @@ def _applyDataset(self):
         was_valid = self.settings_are_valid
         datasetName = self.datasetEdit.text()
         self._datasetNameSlot.setValue(str(datasetName))
-        self.settings_are_valid = str(datasetName) != ""
+        self.settings_are_valid = str(datasetName) != "" or self.default_extension == "zarr"
```
k-dominik, line 77: funny that root array is a thing in zarr. This breaks the container concept in my opinion. But I guess they even kept it for 3.0, so probably I haven't seen the light.

Side note, and not a big issue, but looks like Python's `pathlib` assumes an extension to include the dot (see, e.g. `pathlib.Path.suffix`). If not too much of a big deal I'd recommend changing it accordingly, just to have one thing less to think about.

k-dominik (approved):
Cool, nice refactor, too :) great to unify those things :)


Could've been rebased onto main before opening... (I thought darkmode was broken again...)

volumina/widgets/hierarchicalFileExportOptionsWidget.py
k-dominik: Okay thought more about this empty root array thing. And what I wrote was not true. This will in fact not create a nameless root array. But it is a "logical" leaf node in the ngff tree (I think it should be one at least), so in principle, if all done cleanly this also breaks the hierarchy of ngff-zarr - no other multiscales image can be installed in the root (or at least it shouldn't be possible - although I see, if one would name the scales differently... argh). I'm a bit bummed the ngff spec does not take a firm decision on what the scope of a ngff container should be: multiple multiscale yes (then it should not be allowed to have one at the root), or no (then, they should all be in the root). But whatever, this is not in our hands. Your implementation supports all that is supported by ngff, so that's fine. Or is now the time to get opinionated about this?

btbest: The spec's very first sentence is "The following layout describes the __expected__ Zarr hierarchy for images [...]" (emphasis mine) - implementations can do whatever they like, and they do. E.g. `ngff_zarr` creates a structure of [`/scale0/multiscale_name`](https://github.com/thewtex/ngff-zarr/blob/9305f980a41306efcbd435e025ec715e1dae9990/ngff_zarr/to_multiscales.py#L357). The only requirement is that paths listed in the metadata must actually lead to the respective arrays.

There isn't really a technical or spec issue with allowing one multiscale group at the root, and another at a subpath. The `labels` spec (required layout below) already breaks the rule that "all items in a multiscale group are arrays", though at least it still maintains that "all items in a multiscale group are related to the raw data". Allowing a different multiscale image at a subpath within a root multiscale group would also break the latter.

```
└── image.zarr
    ├── .zgroup
    ├── .zattrs               # multiscales spec
    ├── /scale_0
    ├── /scale_...
    ├── /scale_n
    └── /labels
        ├── .zgroup
        ├── .zattrs          # labels spec
        └── /nuclei
            ├── .zattrs          # multiscales + image-label spec
            ├── /scale_0
            ├── /scale_...
            ├── /scale_n
```

We already have mandatory `datasetName`s for h5, so I wouldn't have a problem with doing the same, but the experience at the recent course (and my personal experience with the FIJI exporter) made it clear that it's confusing for users. I also still have yet to find a use-case for multiple multiscales entries. The only thread on the forum, [from 2022](https://forum.image.sc/t/ome-zarr-storage-structure-for-multiple-multiscale-images/71330), concluded it would be best to make one container per multiscale image. Even the spec [directly suggests](https://ngff.openmicroscopy.org/latest/index.html#image-layout) that separate images should be separate OME-Zarr containers, and the `plate` layout further cements that there ought to be only one multiscale group in a multiscales spec (since it organises plates/wells using lower hierarchy levels).

So given that OME-Zarr tends towards one-dataset-per-container, I would gear ilastik towards the same: No datasetName by default; no merging multiple multiscales into one container (except as labels).

k-dominik: Cool. I think any simplification, or good recommendation we cna make here is useful. I agree having no dataset path as a default seems like a sensible idea (maybe indicating in the dataset name textfield that it is okay to be empty, or hiding it alltogether?).

btbest: I wouldn't hide it altogether since the standard does specify that the multiscale "should" have a name (let's see if I get any [replies](https://forum.image.sc/t/whats-the-use-case-for-multiple-multiscales-in-ome-zarr/103267/4) regarding why)... defaulting to not providing a name is cheeky enough in that context ;)

btbest: Given the replies on the thread, I've concluded that there are probably no existing OME-Zarr stores containing more than one multiscale, and there don't seem to be real-world use-cases for multiple multiscales within one store. Since the only function of the multiscale name is to help the user pick which multiscale from a container they want to look at, the absence of stores with more than one multiscale also makes the name obsolete.

As pointed out in the meeting today, I've concluded that we can safely hide the option from the user altogether. At the same time, I will remove the extra logic for replacing the "exported_data" default value with an empty string in the internal path slot of the op, and instead change the OME-Zarr export on ilastik side to ignore the slot instead (should be less code complexity).

btbest:
Ready to merge when ilastik PR is done

PR #325, opened 2025-06-03
Silence tilescache warning

btbest:
fixmes are not user warnings. This message appears for every user on win and linux (not to mention our test logs), and it's pointless for them.


k-dominik:
how about a nice rebase here to have the tests all green...

k-dominik (approved):

PR #326, opened 2025-06-04
Comments and type hints

btbest:
Minor changes I've been making while investigating slow load times - I figured I'd PR them separately to make review easier (besides the import changes, none of these should be able to change any functionality)


k-dominik (approved):
lgtm, fixing the test with the ilastik env is for another PR

PR #327, opened 2025-06-10
Consistently paint loading tiles white

btbest:
This unifies loading behaviour across applets in ilastik. Previously, applets that only display raw data would leave unloaded tiles invisible, while applets with sparse layers (e.g. user labels) would paint tiles white until data is loaded.

The below example shows the viewer in the data selection applet, where the behaviour now changes. In other applets, e.g. pixel classification Training, the viewer already behaves like in the "After" screenshot.

Unifying the behaviour in the opposite direction (making the "Before" the default everywhere) would be harder, because then volumina would have to inspect layer-tile contents after loading them to determine whether they are empty (and hence whether to paint them). Probably wouldn't be too nasty in terms of performance, but I think the white loading tile is better anyway?

Before:
image

After:
image

### DIFF
--- a/volumina/tiling/tileprovider.py
+++ b/volumina/tiling/tileprovider.py
@@ -322,7 +322,7 @@ class TileProvider(QObject):
             # Don't create white "loading" tiles if there is nothing at all to display
             return None
         qimg = QImage(self.tiling.imageRects[tile_nr].size(), QImage.Format_ARGB32_Premultiplied)
-        qimg.fill(0xFFFFFFFF)
+        qimg.fill(0x00FFFFFF)
         p = QPainter(qimg)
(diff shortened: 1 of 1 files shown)


k-dominik:
Good that you are looking into this. But I can't say I'm a fan. The constant flashing when scrolling (only raw data) is super disturbing. And the white background also works as a "good" placeholder for black-background images (where you'd see a change). But there's also the inverse around where white can be a proper background color and then this is not informative at all. Making it worse everywhere for the sake of consistency is not an approach I want to back.

About introspecting having content or not - I assume a label array with a single pixel annotated would still result in the mostly white background issue... no?

btbest:
I guess there are no easy solutions...

Issue #331, opened 2025-08-13
openGL not working on linux, with certain dependencies (as in the current ilastik development env)


k-dominik:
So far I've observed this only on my ubuntu 20.4 machine.

Creating a current development environment in [ilastik@51138d836a9debac4355077c7b36e001f131dd88](https://github.com/ilastik/ilastik/tree/51138d836a9debac4355077c7b36e001f131dd88).

Repro:

```shell
conda env create -n idev-debug --file ilastik/dev/environment.yaml
conda activate idev-debug
# install volumina
pip install e . 
volumina test.npy --axistags yx
```

with the following npy: [test.zip](https://github.com/user-attachments/files/21750010/test.zip)

which will not show the image and produce a bunch of errors.

k-dominik:
reproduced it on a second machine :/ (Rocky Linux 8.7 (Green Obsidian))

PR #332, opened 2025-08-15
Fix linux not rendering

k-dominik:
partially addresses #331

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -22,7 +22,7 @@ Open an `.npy` image and display it:
 
 ```bash
 # Usage: volumina image axisorder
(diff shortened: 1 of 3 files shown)


k-dominik:
I think this can go in. Side note, this patch was also necessary on the mac after switching to PySide6.

btbest:
#331  now #325 I assume?

k-dominik:
oops yes, of course - I'll edit the original message. Thank you @btbest

k-dominik:
Thank you @btbest! Also good calll with the Readme - fixed!

PR #337, opened 2025-10-02
Speed up config val retrieval by caching

k-dominik:
some properties (verbose_pixelpipeline) are requested a lot of times (like for every PlanarSliceSource request) which adds up. I've seen this being up to 10 percent of the load in the main thread. But I have to admit it's a bit hard to benchmark reliably.

### DIFF
--- a/volumina/config.py
+++ b/volumina/config.py
@@ -22,6 +22,8 @@
 import configparser
 import os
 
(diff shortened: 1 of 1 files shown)


btbest (approved):
I can't say I feel a difference, but also don't see how it would hurt 😅

PR #350, opened 2026-03-15
Fix Screenshot functionality

### LINKED ISSUE #236: Export dialog exports cursor
The volumina export dialog will overlay the current cursor position with the image / overlay
that is being exported. See the red "marker" in this image:

1_sp

I think this is not desired in most use cases. If anyone thinks it might be a useful feature, it could me made optional.

k-dominik:
Screenshot functionality currently is broken in ilastik 1.4.2 betas. This PR fixes it.

Summary:
* newer version combinations of qt/python require tighter types -> casting to int to fix exceptions on opening the dialog, and running the export
* also finally hiding the cursor because I was annoyed
* removed "drawing `mode`" concept from `CrossHairCursor` as it is not used and I don't see a use for it.
* Added some tests after seeing coverage report ;)

fixes: https://github.com/ilastik/volumina/issues/236

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   test-w-ilastik-env:
     strategy:
+      fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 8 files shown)


btbest (approved):
no further comments your honour

PR #356, opened 2026-05-04
Add env variable to distinguish development mode, add dev env yaml

k-dominik:
We used to do special releases of ilastik for stable versions where exceptions that occur during `fetch_layer_tile` would be ignored. These exceptions seem to be triggered by temporary inconsistencies in the graph and are so far transient in nature - no need to inform users about this. But in development mode we do want to see them in order to further identify and reduce such cases.

Environment setup now also sets the environment variable `ILASTIK_DEVELOPMENT_ENV`, which is used to determine if such exceptions should be only logged, or raised.

I also added a development environment.yaml so that the variable is set for development. Also added tests running in the development env.

I considered alternatives, like special releases (which we did in the past), some additional conda-package that could be added to the dev env (which would set the environment var), or .env files. But I found going the proposed route is the easiest with our current dev setup and in principle allows us also to enable this mode in installations if necessary.

Todo after this is merged:
- [x] Follow up in ilastik environment file with the same variable setting: ilastik/ilastik/pull/3212

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,50 @@ jobs:
           fail_ci_if_error: true # optional (default = false)
           verbose: true # optional (default = false)
 
+
+  test-w-dev-env:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, macos-latest, ubuntu-latest]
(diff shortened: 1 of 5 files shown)


btbest (approved):
Works as intended 👍 good doc updates

PR #357, opened 2026-05-04
Fix high dpi UI

btbest:
UI 😭 Idk can't get it to actually look good but at least this way it's usable at all

<details>
  <summary>Before (main)</summary>

  image
</details>

<details>
  <summary>After commit 1</summary>
image
</details>

<details>
  <summary>After commit 2</summary>
image
</details>

<details>
  <summary>After commit 3</summary>
image
</details>

<details>
  <summary>Example from Carving after commit 4</summary>
image
</details>

<details>
  <summary>Before before for comparison (1.4.2b9, without high-dpi, my real desired baseline)</summary>
image
</details>

### DIFF
--- a/volumina/sliceSelectorHud.py
+++ b/volumina/sliceSelectorHud.py
@@ -44,16 +44,33 @@ from qtpy.QtWidgets import (
 from volumina.utility import ShortcutManager
 from volumina.widgets.delayedSpinBox import DelayedSpinBox
 
-TEMPLATE = "QSpinBox {{ color: {0}; font: bold; background-color: {1}; border:0;}}"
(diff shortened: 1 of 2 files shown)


k-dominik:
screenshots show great improvement over the bare highdpi setup!! And I assume the icons (e.g. for slice number) would enlarge and show full numbers...

btbest:
> And I assume the icons (e.g. for slice number) would enlarge and show full numbers...

Ah yes if I make the window slightly wider I do get the full text on the slice selectors :)

k-dominik (approved):
I tried with different scalings and while not perfect it is a _huge_ improvement.. for what it's worth, I like it now better on the mac, too... maybe in the future we'll tune some of these numbers, but I don't want to hold anything back because of that.

PR #358, opened 2026-05-07
Fix high dpi for layer stack

btbest:
Didn't realise this came from volumina :)

* Screenshots show a side-by-side of 1.4.2b9 baseline (left) vs high-dpi (right).
* I'm still working on fixing the splitter in ilastik; it's wide enough to see the layer stack now but apparently I accidentally made that its max width now 😅 
* The "move layer up/down" buttons below the layer stack are from ilastik

<details>
  <summary>Before (main)</summary>

  image
</details>

<details>
  <summary>After</summary>

  image
</details>

### DIFF
--- a/volumina/utility/gui.py
+++ b/volumina/utility/gui.py
@@ -19,6 +19,8 @@
 # This information is also available on the ilastik web site at:
 # 		   http://ilastik.org/license/
 ###############################################################################
+from typing import Optional, Union
+
 from qtpy.QtCore import Qt, QSize
 from qtpy.QtGui import QPixmap
 from qtpy.QtWidgets import QApplication
@@ -34,6 +36,11 @@ def em():
     return QApplication.instance().fontMetrics().ascent()
 
 
+def line_height():
+    """Like `em` but includes descent (slightly larger). Better for vertical spacing/layout."""
+    return QApplication.instance().fontMetrics().height()
+
+
 def phys(logical):
     """Convert a logical pixel value to physical pixels on this device.
     To be avoided, but sometimes necessary to work around Qt scaling quirks."""
@@ -47,4 +54,21 @@ def scale_pixmap(pixmap: QPixmap, size: QSize):
         Qt.SmoothTransformation,
     )
     pixmap.setDevicePixelRatio(dpr())
+    return pixmap
+
+
+def get_responsive_pixmap(*args, logical_size: Optional[Union[QSize, int]] = None):
+    """
+    Load a pixmap with sane HiDPI behavior.
+    args: as passed to QPixmap
+    logical_size: Desired displayed size in logical pixels.
+    """
+    pixmap = QPixmap(*args)
+
+    if logical_size is not None:
+        if isinstance(logical_size, int):
+            logical_size = QSize(logical_size, logical_size)
+
+        pixmap = scale_pixmap(pixmap, logical_size)
+
     return pixmap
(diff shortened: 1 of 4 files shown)


btbest:
Update:

image

The individual layer items still look more compact because the internal spacing is smaller (traded off for equal paddings on top and bottom). Their total height is only a hair smaller than before.

That just sort of happened, but I actually like it more this way...

k-dominik:
It's also a nice improvement on the mac:

image

Only the eye Icons, are too small now for my taste, it makes clicking them much harder now (something that I've seen windows users struggle with before, so I guess good that it works there)...

k-dominik:
The size of the icon I find okay, but it would be good to have a bigger hit box, something like the pink area...:

image

btbest:
Unfortunately that space is reserved for the channel selector :) I made it slightly bigger yet and made it a proper button, so now a larger area should be clickable. This throws out the compactness, but eh...

image

volumina/utility/gui.py
```diff
@@ -0,0 +1,57 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# See the files LICENSE.lgpl2 and LICENSE.lgpl3 for full text of the
+# GNU Lesser General Public License version 2.1 and 3 respectively.
+# This information is also available on the ilastik web site at:
+# 		   http://ilastik.org/license/
+###############################################################################
+from typing import Optional, Union
```
k-dominik:
```suggestion
```

not accessed

volumina/utility/gui.py
```diff
@@ -0,0 +1,57 @@
...
+
+from qtpy.QtCore import Qt, QSize
+from qtpy.QtGui import QPixmap
+from qtpy.QtWidgets import QApplication
+
+
+def dpr():
```
k-dominik: Type annotations here (and for the other functions in this file except scale_pixmap) would be stellar. With newer versions of qt the types you can put into functions become more narrow and it would be great to find these issues before running. At least in my LSP setup I don't get resolved return types for these functions.

volumina/widgets/layerwidget.py
```diff
@@ -124,40 +148,41 @@ def _fractionFromPosition(self, pointf):
...
+        self._eye_open_icon = QIcon(":icons/icons/stock-eye-20.png")
+        self._eye_closed_icon = QIcon(":icons/icons/stock-eye-20-gray.png")
 
-    def active(self):
-        return self._active
+        self.setCheckable(True)
+        self.setChecked(True)
+        self.setAutoRaise(True)
+        self.setToolButtonStyle(Qt.ToolButtonIconOnly)
+        icon_size = round(line_height() * 1.25)
```
k-dominik: What do you think about moving all of these magic constants (1.25 here) to class constants (as you did for `FractionSelectionBar`. I think it would be nice to have some consistency.

k-dominik (approved):
I really like the central place for sensible sizes you added with `gui.py`. I can foresee quite some usage for it.
Also nice that the new `ToggleEye` works with less convoluted logic...
The eye icons are back to normal size for me and clicking them is not to hard. I forgot about the channel selector - too bad, I almost liked the idea of being able to see the icon change while clicking. But with the size restored it's somewhat noticable.

btbest:
I also got rid of one of the relative size helpers as there's only really a need for one. On my display `ascent` is 11px and `height` is 13px for example. Just had to retouch some of the constants a bit relative to the larger base.

btbest:
Reintroduced the right-padding for the layer items... I hadn't noticed the little gap on the right side had disappeared (between the alpha number/opacity bar and the element edge), because I always had it with a scroll bar, which adds its own small spacing.

image

PR #359, opened 2026-05-13
Fix black toggle-eye background on Ubuntu

btbest:
Black button shows up for visible layers on Ubuntu (second layer in the examples)

Current: 
image

PR:
image

On Win the look is unchanged:
image


k-dominik (approved):
lgtm

PR #363, opened 2026-06-19
Python313 fixes

k-dominik:
Okay really not much that popped up during automatic/manual testing, hope that's it :)

the tests with the ilastik env are bit chicken-egg, they are expected to fail at this point (as ilastik-core package needs to be on ilastik-forge...)


btbest:
Seems to be something messed up with my user on GitHub side. Some PR pages, including this one, are only partially loading when I'm logged in.

The parts that don't load unfortunately include the "Add your review" bit, so I can't actually post a review.

So... consider this approved I guess

btbest (approved):
The page is finally loading :D

